PR #2, opened 2022-12-15
Teach open-pr.yml about special pkgver constructs

### LINKED ISSUE #1: Teach open-pr.yml about special pkgver constructs
Some packages like `bash` and `openssl` don't directly set `pkgver` like the workflow currently expects. Teach it to use little update scripts for those as well. This is related to https://github.com/git-for-windows/git/issues/4164

rimrul:
Some packages set `pkgver` in a way that doesn't quite work with our generic `sed` invocation. Teach our `open-pr` workflow to use a package specific update script for that step like we already do with the `updpkgsums` step.

This fixes #1

### DIFF
--- a/.github/workflows/open-pr.yml
+++ b/.github/workflows/open-pr.yml
@@ -152,7 +152,7 @@ jobs:
             exit 0
           fi &&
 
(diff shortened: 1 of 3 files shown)


dscho (approved):
Your work is really helpful @rimrul. Thank you so much for that!

PR #3, opened 2022-12-19
build-and-deploy: fix the run-name for MSYS builds

dscho:
Seems that the `${{ ... }}` construct is not recursive: embedding it within itself won't have the inner one expanded. In other words, [triggering an `i686` build of the `mintty` package resulted in this very ugly title](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3730919260):

	Build and deploy mintty (${{ inputs.architecture }})

Let's avoid that by using [the `format()` function](https://docs.github.com/en/actions/learn-github-actions/expressions#format).


rimrul:
Why did we have that nested structure to begin with? To prevent empty parantheses when  `inputs.architecture` is unset? Is that an expected situation?

dscho:
> Why did we have that nested structure to begin with? To prevent empty parantheses when `inputs.architecture` is unset? Is that an expected situation?

Indeed, to prevent empty parentheses, and it is quite common for `inputs.architecture` to be unset: that's the case for all MINGW builds.

rimrul (approved):

PR #4, opened 2022-12-19
build-and-deploy: add support for arm64 builds

dennisameling:
Note that this only works for self-hosted runners. GitHub Actions will pick up and use the first available arm64 runner.

Here's a successful run for `mingw-w64-openssl` on `aarch64`: https://github.com/dennisameling/git-for-windows-automation/actions/runs/3735260060/jobs/6338362688. Note that I had to do some manual fixes as the pipeline assumes it has a GitHub app (`GH_APP_ID` and `GH_APP_PRIVATE_KEY`) at its disposal. In my run, I simply replaced that with a PAT by doing `const accessToken = ${{ secrets.DENNIS_PAT }}`

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -17,7 +17,7 @@ on:
         description: 'The ref containing the package definition'
         required: true
       architecture:
-        description: 'The architecture to build for (only for MSYS packages)'
+        description: 'The architecture to build for (only for MSYS packages and all arm64 builds)'
         required: false
       actor:
         description: The GitHub user on whose behalf this workflow is run
@@ -37,7 +37,7 @@ env:
 
 jobs:
   build:
-    runs-on: windows-latest
+    runs-on: ${{ github.event.inputs.architecture == 'aarch64' && fromJSON('["Windows", "ARM64"]') || 'windows-latest' }}
     steps:
       - uses: actions/checkout@v3
 
@@ -135,6 +135,8 @@ jobs:
           flavor: ${{ env.PACKAGE_TO_BUILD == 'mingw-w64-git' && 'build-installers' || 'full' }}
           architecture: ${{ env.ARCHITECTURE || 'x86_64' }}
           msys: ${{ env.REPO == 'MSYS2-packages' || env.PACKAGE_TO_BUILD == 'git-for-windows-keyring' }}
+          # We only have to clean up on self-hosted runners
+          cleanup: ${{ runner.arch == 'ARM64' && true || false }}
 
       - name: Clone ${{ env.REPO }}
         shell: bash
@@ -209,8 +211,10 @@ jobs:
             printf '#!/bin/sh\n\nexec /mingw64/bin/git.exe "$@"\n' >/usr/bin/git
           } &&
 
+          MINGW_ARCHS_TO_BUILD=$(test "$ARCHITECTURE" == "aarch64" && echo "clangarm64" || echo "mingw32 mingw64")
+
           cd "/usr/src/$REPO/$PACKAGE_TO_BUILD" &&
-          MAKEFLAGS=-j6 PKGEXT='.pkg.tar.xz' MINGW_ARCH="mingw32 mingw64" $MAKEPKG -s --noconfirm &&
+          MAKEFLAGS=-j6 PKGEXT='.pkg.tar.xz' MINGW_ARCH=$MINGW_ARCHS_TO_BUILD $MAKEPKG -s --noconfirm &&
           cp *.pkg.tar* "$dir/" &&
 
           MAKEFLAGS=-j6 SRCEXT='.src.tar.gz' MINGW_ARCH=mingw64 $MAKEPKG --allsource &&


.github/workflows/build-and-deploy.yml
```diff
@@ -37,7 +37,7 @@ env:
 
 jobs:
   build:
-    runs-on: windows-latest
+    runs-on: ${{ github.event.inputs.architecture == 'aarch64' && fromJSON('["Windows", "ARM64"]') || 'windows-latest' }}
```
dennisameling, line 40: I couldn't use `env.ARCHITECTURE` here so had to resort to `github.event.inputs.architecture`.

.github/workflows/build-and-deploy.yml
```diff
@@ -135,6 +135,8 @@ jobs:
           flavor: ${{ env.PACKAGE_TO_BUILD == 'mingw-w64-git' && 'build-installers' || 'full' }}
           architecture: ${{ env.ARCHITECTURE || 'x86_64' }}
           msys: ${{ env.REPO == 'MSYS2-packages' || env.PACKAGE_TO_BUILD == 'git-for-windows-keyring' }}
+          # We only have to clean up on self-hosted runners
+          cleanup: ${{ (runner.arch == 'ARM64' && true) || false }}
```
dscho, line 139: I had a look at https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context to see whether there would be _any_ indication of a self-hosted runner, but did not find any. Pity.

If only we could detect whether we're running on a self-hosted, _non-ephemeral_ runner, and set `cleanup: true` only then...

Would you mind losing the parentheses? We use the idiom 'A && B || C' (where `&&`/`||` follow [C++'s operator precedence order](https://en.cppreference.com/w/c/language/operator_precedence)) quite often, without parentheses, so that it looks inconsistent when we use them here.

.github/workflows/build-and-deploy.yml
dscho: Right, the environment requires the runner to already run, and that requires `runs-on` to be specified ;-)

dscho (approved):
Very nice! I have just one minor request (dropping some unnecessary parentheses) and will be happy to merge this very soon.

.github/workflows/build-and-deploy.yml
dennisameling: > Would you mind losing the parentheses? 

Done ✅ thanks for sharing the link to that documentation. I guess I got used to the [`no-mixed-operators`](https://eslint.org/docs/latest/rules/no-mixed-operators) rule in ESLint, which encourages using the parentheses to "clarfiy the developer’s intention" 😅

dscho: Ah, that's where that comes from. Yes, I am an unfan of that rule ;-)

.github/workflows/build-and-deploy.yml
```diff
@@ -209,8 +211,10 @@ jobs:
             printf '#!/bin/sh\n\nexec /mingw64/bin/git.exe "$@"\n' >/usr/bin/git
           } &&
 
+          MINGW_ARCHS_TO_BUILD=$(test "$ARCHITECTURE" == "aarch64" && echo "clangarm64" || echo "mingw32 mingw64")
```
dscho, line 214: @dennisameling I hope you don't mind that I edited this: We can either use a Bash-style ternary (with `$ARCHITECTURE`) or use GitHub workflow syntax (`${{ env.ARCHITECTURE == 'aarch64' && 'clangarm64' || 'mingw32 mingw64' }}`), but we should not mix them. I opted for the former.

dscho (approved):

dscho:
image

Now I'm going to start the VM.

dscho:
[So exciting!!!](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3760961637/jobs/6392212187)

image

.github/workflows/build-and-deploy.yml
dennisameling: Works for me! 👍🏼

dennisameling:
Yay!!!! So exciting indeed! 🚀

dscho:
Ah... [It failed because `pacman` needed to be upgraded](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3760961637/jobs/6392212187#step:10:51). Let me see whether I can easily fix the `sync` workflow (running it manually did not work because [the job was skipped](https://github.com/git-for-windows/git-sdk-arm64/actions/runs/3761014987/jobs/6392312970)).

dscho:
> Let me see whether I can easily fix the `sync` workflow

[Finally](https://github.com/git-for-windows/git-sdk-arm64/compare/5049de3fbc52790a8b5cb0e21fd1f0b9cde5fe28...b6a863365fa7c22c91d7dbfd993f404da9a2b559) it [worked](https://github.com/git-for-windows/git-sdk-arm64/actions/runs/3761904306).

dscho:
What did not work was the [GPG step](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3760961637/jobs/6394127463#step:12:31). I will have to leave it at that, for now.

dennisameling:
Oh my 😅 that commit history looks really similar to my [mess](https://github.com/dennisameling/git/commits/clangarm64-tmp-test-dennis) from testing yesterday haha. Thank you _so_ much for going through that! I know it can be a total pain!

Let's say if I had some time next week and wanted to kick off the [build-and-deploy](https://github.com/git-for-windows/git-for-windows-automation/blob/main/.github/workflows/build-and-deploy.yml) pipeline for arm64 to build some of the [packages](https://github.com/dennisameling/git/commit/abcdc80caee9ca0dd22835c6d6dd9d83f71620cc#diff-027c70cd6ed99ca64ca6fc12148edaf142051f780a5516aca15ef28f437b711aR356-R362) (`git-lfs`, `git-credential-manager`, etc.) for the `git-sdk-arm64`, how would I go about that? I would either need:
- Access to add (temporary) self-hosted runners to `git-for-windows-automation` (I currently can't do that, see screenshot below), so that I can at least add some initial package versions for those to `git-sdk-arm64`
- Your permission to build those packages locally on my machine (with my code signing + GPG signing) and commit them to `git-sdk-arm64`.

WDYT?

image

dennisameling:
> What did not work was the [GPG step](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3760961637/jobs/6394127463#step:12:31). I will have to leave it at that, for now.

Fixed in https://github.com/git-for-windows/git-for-windows-automation/pull/8 👍🏼

dscho:
The fruit of all of your labor: https://wingit.blob.core.windows.net/aarch64/git-for-windows-aarch64.db

PR #7, opened 2022-12-24
build-and-deploy: use separate check runs for each MSys architecture

rimrul:
For MSys packages we start two parrallel runs of this workflows, that can succeed or fail independently from one another. We need to use two separate check runs to accurately represent their status. The corresponding change in gfw-helper-github-app is part of https://github.com/git-for-windows/gfw-helper-github-app/pull/17

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -90,7 +90,11 @@ jobs:
               process.env.OWNER,
               process.env.REPO,
               process.env.REF,
-              process.env.BUILD_ONLY ? 'build' : 'deploy',
+              `${
+                process.env.BUILD_ONLY ? 'build' : 'deploy'
+              }${
+                process.env.ARCHITECTURE ? `_${process.env.ARCHITECTURE}` : ''
+              }`,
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -90,7 +90,7 @@ jobs:
               process.env.OWNER,
               process.env.REPO,
               process.env.REF,
-              process.env.BUILD_ONLY ? 'build' : 'deploy',
+              (process.env.BUILD_ONLY ? 'build' : 'deploy') + (process.env.REPO=='MSYS2-packages' && process.env.ARCHITECTURE != '' ? '' : '_' + process.env.ARCHITECTURE),
```
dscho, line 93: I think we also need this for the upcoming ARM64 builds:

```suggestion
              `${
                process.env.BUILD_ONLY ? 'build' : 'deploy'
              }${
                process.env.ARCHITECTURE ? `_${process.env.ARCHITECTURE}` : ''
              }`,
```

rimrul: I think you're right. At first I thought we only needed this for Msys-Packages, because we currently only need one workflow run for Mingw-Packages, but that's about to change.

dscho (approved):
@rimrul thank you so much for this PR!

PR #11, opened 2023-01-02
Optionally keep self-hosted-runners allocated after creating them

dscho:
For now, we manually spin up the runners. It makes more sense to keep them running so that they can pick up the job immediately instead of deallocating them and then manually restart them.

### DIFF
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -27,7 +27,7 @@ on:
         type: string
         required: true
         description: Deallocate the runner immediately after creating it (useful for spinning up runners preemptively)
-        default: true
+        default: false
 
 env:
   AMOUNT_OF_RUNNERS: ${{ github.event.inputs.amount_of_runners }}
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -37,6 +37,9 @@ env:
   DEALLOCATE_IMMEDIATELY: ${{ github.event.inputs.deallocate_immediately }}
   # This has to be a public URL that the VM can access after creation
   POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref }}/azure-self-hosted-runners/post-deployment-script.ps1
+  # Note that you'll need "p" (arm64 processor) and ideally "d" (local temp disk). The number 8 stands for 8 CPU-cores.
+  # For a convenient overview of all arm64 VM types, see e.g. https://azureprice.net/?_cpuArchitecture=Arm64
+  AZURE_VM_TYPE: Standard_D8plds_v5
 
 # The following secrets are required for this workflow to run:
 # AZURE_CREDENTIALS - Credentials for the Azure CLI. It's recommended to set up a resource
@@ -141,7 +144,7 @@ jobs:
       with:
         resourceGroupName: ${{ secrets.AZURE_RESOURCE_GROUP }}
         template: ./azure-self-hosted-runners/azure-arm-template.json
-        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize=Standard_D8pls_v5 publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }} stopService=${{ env.DEALLOCATE_IMMEDIATELY }}
+        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize="${{ env.AZURE_VM_TYPE }}" publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }} stopService=${{ env.DEALLOCATE_IMMEDIATELY }}
 
     - name: Deallocate the VM for later use
       if: env.DEALLOCATE_IMMEDIATELY == 'true'
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -23,12 +23,18 @@ on:
         type: string
         required: false
         description: Repo to deploy the runner to. Only needed if runner_scope is set to "repo-level" (defaults to current repository)
+      deallocate_immediately:
+        type: string
+        required: true
+        description: Deallocate the runner immediately after creating it (useful for spinning up runners preemptively)
+        default: true
 
 env:
   AMOUNT_OF_RUNNERS: ${{ github.event.inputs.amount_of_runners }}
   ACTIONS_RUNNER_SCOPE: ${{ github.event.inputs.runner_scope }}
   ACTIONS_RUNNER_ORG: "${{ github.event.inputs.runner_org || github.repository_owner }}"
   ACTIONS_RUNNER_REPO: "${{ github.event.inputs.runner_repo || github.event.repository.name }}"
+  DEALLOCATE_IMMEDIATELY: ${{ github.event.inputs.deallocate_immediately }}
   # This has to be a public URL that the VM can access after creation
   POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref }}/azure-self-hosted-runners/post-deployment-script.ps1
 
@@ -135,9 +141,10 @@ jobs:
       with:
         resourceGroupName: ${{ secrets.AZURE_RESOURCE_GROUP }}
         template: ./azure-self-hosted-runners/azure-arm-template.json
-        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize=Standard_D8pls_v5 publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }}
+        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize=Standard_D8pls_v5 publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }} stopService=${{ env.DEALLOCATE_IMMEDIATELY }}
 
     - name: Deallocate the VM for later use
+      if: env.DEALLOCATE_IMMEDIATELY == 'true'
       uses: azure/CLI@v1
       with:
         azcliversion: 2.43.0
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -134,6 +134,15 @@ jobs:
           | jq --raw-output .token)
         echo "::add-mask::$ACTIONS_RUNNER_TOKEN"
         echo ACTIONS_RUNNER_TOKEN=$ACTIONS_RUNNER_TOKEN >> $GITHUB_ENV
+
+        # The Azure VM type we use has blazing-fast local, temporary storage available as the D:\ drive.
+        # The only downside is that, after dellocation, the contents of this disk (including the Actions Runner),
+        # are destroyed. Let's only use it when we don't immediately deallocate the VM.
+        if [[ "${{ env.DEALLOCATE_IMMEDIATELY }}" == "true" ]]; then
+          echo "ACTIONS_RUNNER_PATH=C:\a" >> $GITHUB_ENV
+        else
+          echo "ACTIONS_RUNNER_PATH=D:\a" >> $GITHUB_ENV
+        fi
         
     - name: Azure Login
       uses: azure/login@v1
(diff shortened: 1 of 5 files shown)


dscho:
I tested this in conjunction with https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3825495631/jobs/6508527516 and was surprised that the runner did not come online immediately... turns out that I had forgotten about the post-deploy script! I'm not so much of a PowerShell expert, so I want to use the next deployment of an aarch64 package as an excuse to test this before I mark this PR as ready for review.

dennisameling (commented):
@dscho I updated two things in your code:

In the `.yml` file, I replaced `stopService=$env:DEALLOCATE_IMMEDIATELY` with `stopService={{ env.DEALLOCATE_IMMEDIATELY }}`.

In the `azure-arm-template.json` file, I added the `stopService` parameter to the `parameters` object on top of the file. It needs to be defined there before you can actually use it as `parameters('stopService')`.

I also took the opportunity to use the `D:\` drive if possible for faster Actions runs. Hope that's OK!

dscho:
> In the `.yml` file, I replaced `stopService=$env:DEALLOCATE_IMMEDIATELY` with `stopService={{ env.DEALLOCATE_IMMEDIATELY }}`.

I think there is a `$` missing before the `{{`.

But I have a more fundamental concern with using `${{ env.XYZ }}` instead of `$env:XYZ`: the latter is a lot more robust with respect to spaces, quotes and other special characters in the value of the environment variable. In general, I think we should always use the native environment variable expansion in scripts, i.e. `$env:XYZ` in PowerShell, `$XYZ` in Bash.

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -135,9 +141,10 @@ jobs:
...
-        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize=Standard_D8pls_v5 publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }}
+        parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json githubActionsRunnerRegistrationUrl="${{ env.ACTIONS_RUNNER_REGISTRATION_URL }}" githubActionsRunnerToken="${{ env.ACTIONS_RUNNER_TOKEN }}" postDeploymentPsScriptUrl="${{ env.POST_DEPLOYMENT_SCRIPT_URL }}" virtualMachineName=${{ steps.generate-vm-name.outputs.vm_name }} virtualMachineSize=Standard_D8pls_v5 publicIpAddressName1=${{ steps.generate-vm-name.outputs.vm_name }}-ip adminUsername=${{ secrets.AZURE_VM_USERNAME }} adminPassword=${{ secrets.AZURE_VM_PASSWORD }} stopService={{ env.DEALLOCATE_IMMEDIATELY }}
```
dscho: This is where the `{{` needs to be replaced with `${{`. But as I said, I'd like to avoid that construct in favor of native PowerShell environment variable expansion here (and elsewhere in this part of the workflow definition, in a separate commit).

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -37,6 +37,9 @@ env:
   DEALLOCATE_IMMEDIATELY: ${{ github.event.inputs.deallocate_immediately }}
   # This has to be a public URL that the VM can access after creation
   POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref_name }}/azure-self-hosted-runners/post-deployment-script.ps1
+  # Note that you'll need "p" (arm64 processor) and ideally "l" (local temp disk). The number 8 stands for 8 CPU-cores.
+  # For a convenient overview of all arm64 VM types, see e.g. https://azureprice.net/?_cpuArchitecture=Arm64
+  AZURE_VM_TYPE: Standard_D8plds_v5
```
dscho, line 42: Nice thinking @dennisameling !

Could you add another knob where the Azure region can be configured, defaulting to "US West 2" (which seems to be more cost-effective than "West Europe", which is used currently)?

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -134,6 +134,15 @@ jobs:
           | jq --raw-output .token)
         echo "::add-mask::$ACTIONS_RUNNER_TOKEN"
         echo ACTIONS_RUNNER_TOKEN=$ACTIONS_RUNNER_TOKEN >> $GITHUB_ENV
+
+        # The Azure VM type we use has blazing-fast local, temporary storage available as the D:\ drive.
+        # The only downside is that, after dellocation, the contents of this disk (including the Actions Runner),
```
dscho, line 139: That'd make for a fine reason to switch the default to _not_ deallocate by default, don't you think @dennisameling ?

dscho (commented):
Great work!

.github/workflows/create-azure-self-hosted-runners.yml
dennisameling: > This is where the `{{` needs to be replaced with `${{`. 

Oops! Good catch. Will fix.

> But as I said, I'd like to avoid that construct in favor of native PowerShell environment variable expansion here (and elsewhere in this part of the workflow definition, in a separate commit).

Unfortunately, that won't work. The `DEALLOCATE_IMMEDIATELY` env variable is only available in the Actions Runner context. When the VM is created, it does the following:

1. Download the post-deployment script from https://raw.githubusercontent.com/git-for-windows/git-for-windows-automation/main/azure-self-hosted-runners/post-deployment-script.ps1
2. Execute it _from within the VM_ (which doesn't have env variables like `DEALLOCATE_IMMEDIATELY`)

So we need to explicitly pass these parameters with their actual values to the `azure/arm-deploy` action, which in turn [passes it](https://github.com/git-for-windows/git-for-windows-automation/blob/37df492868dcb7f01ff8682e03d66a4badbfd853/azure-self-hosted-runners/azure-arm-template.json#L233-L260) to the `Microsoft.CustomScriptExtension`, which in turn will execute the two steps above.

dennisameling: Sorry, I think I misunderstood actually:

> In general, I think we should always use the native environment variable expansion in scripts, i.e. `$env:XYZ` in PowerShell, `$XYZ` in Bash.

That makes sense to me. Just replacing `${{ env.XYZ }}` _everywhere_ in the workflow. Will do!

.github/workflows/create-azure-self-hosted-runners.yml
dennisameling: Will do! Fun fact: it looks like "Central India" is ~36% cheaper than US-based VMs. Shall I default to that instead?

https://azureprice.net/vm/Standard_D8plds_v5

.github/workflows/create-azure-self-hosted-runners.yml
dennisameling: Makes sense to me! Once we have the logic in place to automatically remove a runner when a job is finished (as you mentioned [here](https://github.com/git-for-windows/git/discussions/4186#discussioncomment-4576493)), and when we have logic in place that creates those runners on-demand when needed, we shouldn't have to deallocate things, but just live with the few mins of setup time for those jobs.

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -23,14 +23,23 @@ on:
...
 
 env:
   AMOUNT_OF_RUNNERS: ${{ github.event.inputs.amount_of_runners }}
   ACTIONS_RUNNER_SCOPE: ${{ github.event.inputs.runner_scope }}
   ACTIONS_RUNNER_ORG: "${{ github.event.inputs.runner_org || github.repository_owner }}"
   ACTIONS_RUNNER_REPO: "${{ github.event.inputs.runner_repo || github.event.repository.name }}"
+  DEALLOCATE_IMMEDIATELY: ${{ github.event.inputs.deallocate_immediately }}
   # This has to be a public URL that the VM can access after creation
   POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref_name }}/azure-self-hosted-runners/post-deployment-script.ps1
+  # Note that you'll need "p" (arm64 processor) and ideally "l" (local temp disk). The number 8 stands for 8 CPU-cores.
```
rimrul, line 40: According to https://azureprice.net/vm/Standard_D8plds_v5 the `d` is whats responsible for the temp disk.

> p – Arm64 processor
d – Diskfull (local temp disk is present)
l – A lower amount of memory than the memory intensive size

dennisameling: Good catch, thank you! Will fix.

.github/workflows/create-azure-self-hosted-runners.yml
dscho: > Makes sense to me! Once we have the logic in place to automatically remove a runner when a job is finished (as you mentioned [here](https://github.com/git-for-windows/git/discussions/4186#discussioncomment-4576493)), and when we have logic in place that creates those runners on-demand when needed, we shouldn't have to deallocate things, but just live with the few mins of setup time for those jobs.

Right. And I could imagine that hosted ARM64-based runners will come at some point, anyway. So those few minutes won't be needed forever.

.github/workflows/create-azure-self-hosted-runners.yml
dscho: > Will do! Fun fact: it looks like "Central India" is ~16% cheaper than US-based VMs. Shall I default to that instead?
> 
> https://azureprice.net/vm/Standard_D8plds_v5

Good point, let's do that!

.github/workflows/create-azure-self-hosted-runners.yml
dennisameling: How does [this](https://github.com/git-for-windows/git-for-windows-automation/pull/11/commits/52bab03c5999115facfb5da8685b682b574f4be8) look?

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -40,6 +40,8 @@ env:
   # Note that you'll need "p" (arm64 processor) and ideally "d" (local temp disk). The number 8 stands for 8 CPU-cores.
   # For a convenient overview of all arm64 VM types, see e.g. https://azureprice.net/?_cpuArchitecture=Arm64
   AZURE_VM_TYPE: Standard_D8plds_v5
+  # At the time of writing, "centralindia" was the cheapest region for the VM type we're using. https://azureprice.net/vm/Standard_D8plds_v5
+  AZURE_VM_REGION: centralindia
```
rimrul, line 44: The commit message says `westindia`, while `centralindia` seems to be correct.

dennisameling: Ugh 🤦🏼 thanks for being so sharp on this! My brain is not fully awake yet. Will fix

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -124,7 +135,32 @@ jobs:
...
+          publicIpAddressName1="${{ steps.generate-vm-name.outputs.vm_name }}-ip"
+          adminUsername="${{ secrets.AZURE_VM_USERNAME }}"
+          adminPassword="${{ secrets.AZURE_VM_PASSWORD }}"
+          stopService="$DEALLOCATE_IMMEDIATELY"
+          githubActionsRunnerPath="$ACTIONS_RUNNER_PATH"
+          location="$AZURE_VM_REGION"
+        END
+        )
+
+        echo "AZURE_ARM_PARAMETERS=$AZURE_ARM_PARAMETERS" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
dscho: Ah, you need a `<<EOF` trick: see e.g. [this blog post](https://medium.com/@ibraheemabukaff/github-actions-exporting-multi-line-one-line-value-environment-variable-5bb86d01e866).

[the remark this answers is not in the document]
dennisameling: I tested it locally, and `AZURE_ARM_PARAMETERS=$(cat <<-END` actually turns it into a single-line, space-separated string, so then it can be written into the `$GITHUB_ENV` without having to do the `<<EOF` trick you mentioned 🙏🏼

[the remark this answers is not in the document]
dscho: Ah, I missed that the `$(...)` aren't enclosed in double-quotes. Excellent!

.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -23,14 +19,24 @@ on:
...
   ACTIONS_RUNNER_REPO: "${{ github.event.inputs.runner_repo || github.event.repository.name }}"
+  DEALLOCATE_IMMEDIATELY: ${{ github.event.inputs.deallocate_immediately }}
   # This has to be a public URL that the VM can access after creation
-  POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref_name }}/azure-self-hosted-runners/post-deployment-script.ps1
+  POST_DEPLOYMENT_SCRIPT_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ github.ref }}/azure-self-hosted-runners/post-deployment-script.ps1
+  # Note that you'll need "p" (arm64 processor) and ideally "d" (local temp disk). The number 8 stands for 8 CPU-cores.
+  # For a convenient overview of all arm64 VM types, see e.g. https://azureprice.net/?_cpuArchitecture=Arm64
+  AZURE_VM_TYPE: Standard_D8plds_v5
+  # At the time of writing, "centralindia" was the cheapest region for the VM type we're using. https://azureprice.net/vm/Standard_D8plds_v5
+  AZURE_VM_REGION: centralindia
```
dscho, line 39: Hmm. It does [not look as if that region has the VM type we require](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3883922611/jobs/6625799742#step:7:21).

dennisameling: Interesting. AzurePrice [does list it as being available in that region](https://azureprice.net/vm/Standard_D8plds_v5). Probably "eastus" would do the trick then 🤷🏼

dscho: https://azureprice.net/content/about says that this site is not affiliated with Microsoft, and I fail to see any equivalent prices in the Azure Portal (probably because the former tries to factor in Windows license costs?)

In any case, I think I solved the riddle about West India:

image

Unfortunately, I do not find any official documentation about pricing at https://learn.microsoft.com/en-us/azure/virtual-machines/dplsv5-dpldsv5-series, at least not anything helpful (the calculator seems to either be unaware of that series or calls it by a different name).

dscho: I went with "westus2" and started [a new workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3884683198).

dscho: Oh, and https://azure.microsoft.com/en-us/explore/global-infrastructure/products-by-region/?products=virtual-machines&regions=central-india,south-india,west-india claims that availability of these machines in India is projected for Q1 2023.

dennisameling: Almost there!! 😅 🤦🏼 https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3881130673/jobs/6625764356

```
 ==> Verifying source file signatures with gpg...
    xpdf-4.00.tar.gz ... FAILED (unknown public key F4825F5397271342)
==> ERROR: One or more PGP signatures could not be verified!
```

.github/workflows/delete-self-hosted-runner.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Azure Login
+      uses: azure/login@v1
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+    - name: Delete VM '${{ env.ACTIONS_RUNNER_NAME }}'
+      uses: azure/CLI@v1
+      with:
+        azcliversion: 2.43.0
+        inlineScript: |
+          az vm delete -n "$ACTIONS_RUNNER_NAME" -g ${{ secrets.AZURE_RESOURCE_GROUP }} --yes
```
dscho, line 30: This works, kind of. But it leaves two resources behind, the network security group and the virtual network:

image

@dennisameling you mentioned previously that we could avoid that by reusing a single resource? Or is there a way to pass one or two `deleteOption`s so that these resources are deleted, too, when the VM is deleted?

dscho: Let's see whether 4c2ba65 works.

dennisameling: Looks like this is a known issue: https://github.com/Azure/azure-cli/issues/22158

We might need 1-2 additional `az` CLI calls to find the NSG/Vnet that are attached to the VM and delete them in follow-up calls.

dennisameling: It also looks like `deleteOption` isn't a supported property on `Microsoft.Network/networkSecurityGroups` and `Microsoft.Network/virtualNetworks`:

https://learn.microsoft.com/en-us/azure/templates/microsoft.network/networksecuritygroups?pivots=deployment-language-arm-template#networksecuritygrouppropertiesformat

https://learn.microsoft.com/en-us/azure/templates/microsoft.network/virtualnetworks?pivots=deployment-language-arm-template

dscho: Indeed, [it does not work](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3885243390/jobs/6628872716#step:7:21).

> We might need 1-2 additional `az` CLI calls to find the NSG/Vnet that are attached to the VM and delete them in follow-up calls.

I guess we need 2 additional CLI calls, but we know the exact name i.e. we do not have to identify the resources via a separate call.

.github/workflows/create-azure-self-hosted-runners.yml
dscho: Yep. I actually had that fixed in git-for-windows/MINGW-packages@c8ccd71286c52c2f4df17857a2e8f40830eb20fa already, but then had a brain fart when I re-ran the failed jobs, not remembering that a re-run would fail to target the newly-pushed commit.

.github/workflows/create-azure-self-hosted-runners.yml
dscho: > How does [this](https://github.com/git-for-windows/git-for-windows-automation/pull/11/commits/52bab03c5999115facfb5da8685b682b574f4be8) look?

It looks good! I did have to change the `ACTIONS_RUNNER_REGISTRATION_URL` lines because they only set the environment variable for subsequent steps, not the current one, but we no longer use it in a subsequent step, but in the current one.

dscho:
I'm declaring this ready to be merged, after it successfully created several runners in the attempt to [build `mingw-w64-clang-aarch64-xpdf-tools`](https://github.com/git-for-windows/git-sdk-arm64/issues/3) 😉

dennisameling:
Thanks for your work on this!! 🔥

PR #12, opened 2023-01-09
Prepare for building arm64 versions of the `wintoast` and `git-credential-manager` packages

dscho:
As discussed [here](https://github.com/git-for-windows/git-sdk-arm64/issues/3#issuecomment-1371496963).

PR #14, opened 2023-01-10
build-and-deploy: use hosted runners for git-lfs for ARM64

dscho:
This is similar in spirit to #12, just for Git LFS (which I really should have included but I forgot that it was in the same boat as wintoast/Git Credential Manager).


dscho:
[This workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3887540853) deployed `mingw-w64-clang-aarch64-git-lfs`.

dennisameling (approved):
Thank you!

PR #17, opened 2023-01-12
Use the Github API to obtain the latest Git for Windows and Github Actions Runner releases

### LINKED ISSUE #16: create-self-hosted-runner: automatically use the latest Git for Windows and GitHub Actions runner version
When spinning up the self-hosted Windows/ARM64 runners, we [hard-code the Git for Windows version](https://github.com/git-for-windows/git-for-windows-automation/blob/200fdc9be4b4af6dc8751f2e18202eeab9ff5253/azure-self-hosted-runners/post-deployment-script.ps1#L32-L34) that gets installed, and likewise we [hard-code the GitHub Actions runner version](https://github.com/git-for-windows/git-for-windows-automation/blob/200fdc9be4b4af6dc8751f2e18202eeab9ff5253/azure-self-hosted-runners/post-deployment-script.ps1#L36-L39).

Let's determine the latest versions of both components automatically.

This requires a little bit of PowerShell scripting, accessing https://api.github.com/repos/git-for-windows/git/releases/latest and https://api.github.com/repos/actions/runner/releases/latest. Both the version string as well as the SHA-256 checksums need to be parsed out of the response.

hmartinez82:
Fixes #16

### DIFF
--- a/azure-self-hosted-runners/post-deployment-script.ps1
+++ b/azure-self-hosted-runners/post-deployment-script.ps1
@@ -29,14 +29,25 @@ Write-Output "Starting post-deployment script."
 # TOOL VERSIONS AND OTHER VARIABLES
 # =================================
 
-$GitForWindowsVersion = "2.39.0"
-$GitForWindowsTag = "2.39.0.windows.1"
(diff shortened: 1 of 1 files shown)


hmartinez82:
Closing this for now, since I got the meaning of SHA wrong. We need the sha of the asset that's being downloaded and not the sha of the commit of the release 😅

PR #22, opened 2023-01-22
Port `git-artifacts` pipeline to `git-for-windows-automation`

dennisameling:
Successful run: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/3985052314

### DIFF
--- a/.github/workflows/git-artifacts.yml
+++ b/.github/workflows/git-artifacts.yml
@@ -22,6 +22,7 @@ on:
         options: 
         - x86_64
         - i686
+        - aarch64
       bundle_artifacts_workflow_run_id:
         description: 'Workflow run ID from bundle-artifacts pipeline'
         required: false
@@ -42,7 +43,7 @@ defaults:
 
 jobs:
   pkg:
-    runs-on: windows-latest
+    runs-on: ${{ github.event.inputs.architecture == 'aarch64' && fromJSON('["Windows", "ARM64"]') || 'windows-latest' }}
     outputs:
       artifact_matrix: ${{steps.artifact-build-matrix.outputs.matrix}}
       msystem: ${{steps.configure-environment.outputs.MSYSTEM}}
@@ -64,6 +65,12 @@ jobs:
               MINGW_PREFIX=/mingw32
               MINGW_PACKAGE_PREFIX=mingw-w64-i686
               SDK_REPO_ARCH=32
+              ;;
+            aarch64)
+              MSYSTEM=CLANGARM64
+              MINGW_PREFIX=/clangarm64
+              MINGW_PACKAGE_PREFIX=mingw-w64-clang-aarch64
+              SDK_REPO_ARCH=arm64
               ;;
             *)
               echo "Unhandled architecture: $ARCHITECTURE"
@@ -194,7 +201,7 @@ jobs:
           name: pkg-${{env.ARCHITECTURE}}
           path: artifacts
   artifacts:
-    runs-on: windows-latest
+    runs-on: ${{ github.event.inputs.architecture == 'aarch64' && fromJSON('["Windows", "ARM64"]') || 'windows-latest' }}
     needs: pkg
     env:
       MSYSTEM: ${{ needs.pkg.outputs.msystem }}
--- a/.github/workflows/git-artifacts.yml
+++ b/.github/workflows/git-artifacts.yml
@@ -1,0 +1,265 @@
+name: git-artifacts
+run-name: Build git-artifacts (${{ inputs.architecture }})
+
+on:
+  workflow_dispatch:
+    inputs:
+      artifacts:
+        description: 'Optionally restrict what artifacts to build (portable, installer, etc.). Separate artifacts with spaces'
+        required: false
+      ref:
+        description: 'Optionally override which branch to build'
+        required: false
+        default: main
+      repository:
+        description: 'Optionally override from where to fetch the specified ref'
+        required: false
+        default: git-for-windows/git
+      architecture:
+        type: choice
+        description: 'Architecture to build'
+        required: true
+        options: 
+        - x86_64
+        - i686
+      bundle_artifacts_workflow_run_id:
+        description: 'Workflow run ID from bundle-artifacts pipeline'
+        required: false
+
+env:
+  GPG_OPTIONS: "--batch --yes --no-tty --list-options no-show-photos --verify-options no-show-photos --pinentry-mode loopback"
+  HOME: "${{github.workspace}}\\home"
+  USERPROFILE: "${{github.workspace}}\\home"
+  ARTIFACTS_TO_BUILD: "${{github.event.inputs.artifacts}}"
+  REPOSITORY: "${{github.event.inputs.repository}}"
+  REF: "${{github.event.inputs.ref}}"
+  ARCHITECTURE: "${{github.event.inputs.architecture}}"
+  BUNDLE_ARTIFACTS_WORKFLOW_RUN_ID: "${{github.event.inputs.bundle_artifacts_workflow_run_id}}"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  pkg:
+    runs-on: windows-latest
+    outputs:
+      artifact_matrix: ${{steps.artifact-build-matrix.outputs.matrix}}
+      msystem: ${{steps.configure-environment.outputs.MSYSTEM}}
+      mingw_package_prefix: ${{steps.configure-environment.outputs.MINGW_PACKAGE_PREFIX}}
+      sdk_repo_arch: ${{steps.configure-environment.outputs.SDK_REPO_ARCH}}
+    steps:
+      - name: Configure environment
+        id: configure-environment
+        run: |
+          case "$ARCHITECTURE" in
+            x86_64)
+              MSYSTEM=MINGW64
+              MINGW_PREFIX=/mingw64
+              MINGW_PACKAGE_PREFIX=mingw-w64-x86_64
+              SDK_REPO_ARCH=64
+              ;;
+            i686)
+              MSYSTEM=MINGW32
+              MINGW_PREFIX=/mingw32
+              MINGW_PACKAGE_PREFIX=mingw-w64-i686
+              SDK_REPO_ARCH=32
+              ;;
+            *)
+              echo "Unhandled architecture: $ARCHITECTURE"
+              exit 1
+              ;;
+          esac
+          echo "MSYSTEM=$MSYSTEM" >> $GITHUB_ENV
+          echo "MSYSTEM=$MSYSTEM" >> $GITHUB_OUTPUT
+          echo "MINGW_PREFIX=$MINGW_PREFIX" >> $GITHUB_ENV
+          echo "MINGW_PACKAGE_PREFIX=$MINGW_PACKAGE_PREFIX" >> $GITHUB_ENV
+          echo "MINGW_PACKAGE_PREFIX=$MINGW_PACKAGE_PREFIX" >> $GITHUB_OUTPUT
+          echo "SDK_REPO_ARCH=$SDK_REPO_ARCH" >> $GITHUB_OUTPUT
+          echo "ARTIFACTS_TO_BUILD=$(test -n "$ARTIFACTS_TO_BUILD" && echo $ARTIFACTS_TO_BUILD || \
+            echo "installer portable archive mingit$(test "$ARCHITECTURE" = aarch64 || echo ' mingit-busybox')")" >> $GITHUB_ENV
+      - name: Configure user
+        run:
+          USER_NAME="${{github.actor}}" &&
+          USER_EMAIL="${{github.actor}}@users.noreply.github.com" &&
+          mkdir "$HOME" &&
+          git config --global user.name "$USER_NAME" &&
+          git config --global user.email "$USER_EMAIL" &&
+          echo "PACKAGER=$USER_NAME <$USER_EMAIL>" >> $GITHUB_ENV
+      - uses: git-for-windows/setup-git-for-windows-sdk@v1
+        with:
+          flavor: build-installers
+          architecture: ${{env.architecture}}
+      - name: clone git-for-windows-automation
+        run: git clone --single-branch -b $GITHUB_REF_NAME https://github.com/git-for-windows/git-for-windows-automation
+      - uses: actions/github-script@v6
+        id: artifact-build-matrix
+        name: Create artifact build matrix
+        with:
+          script: |
+            core.info('Preparing artifact build matrix...')
+            const createArtifactsMatrix = require('./git-for-windows-automation/create-artifacts-matrix')
+            createArtifactsMatrix(core, process.env.ARTIFACTS_TO_BUILD, process.env.ARCHITECTURE)
+      - uses: actions/github-script@v6
+        id: get-bundle-artifacts-url
+        name: Get bundle-artifacts download URL
+        with:
+          script: |
+            const getArtifact = require('./git-for-windows-automation/get-bundle-artifacts-artifact')
+            const workflowId = process.env.BUNDLE_ARTIFACTS_WORKFLOW_RUN_ID
+            core.info('Getting download URL for bundle-artifacts...')
+            await getArtifact(github, context, core, workflowId)
+      - name: Download bundle-artifacts zip
+        run: |
+          mkdir bundle-artifacts
+          curl -o bundle-artifacts.zip "${{steps.get-bundle-artifacts-url.outputs.downloadUrl}}"
+          unzip bundle-artifacts.zip -d bundle-artifacts
+          echo "GIT_VERSION=$(cat bundle-artifacts/next_version)" >> $GITHUB_ENV
+      - name: Re-publish bundle-artifacts so the next job can easily use it
+        uses: actions/upload-artifact@v3
+        with:
+          name: bundle-artifacts
+          path: bundle-artifacts
+      - name: Clone and update build-extra
+        run: |
+          d=/usr/src/build-extra &&
+          if test ! -d $d/.git
+          then
+            git clone --single-branch -b main https://github.com/git-for-windows/build-extra $d
+          else
+            git -C $d fetch https://github.com/git-for-windows/build-extra main &&
+            git -C $d switch -C main FETCH_HEAD
+          fi &&
+          git -C $d pull "$PWD"/bundle-artifacts/build-extra.bundle main
+      - name: Check out git/git
+        shell: bash
+        run: |
+          git -c init.defaultBranch=main init &&
+          git remote add -f origin https://github.com/git-for-windows/git &&
+          git fetch --tags bundle-artifacts/git.bundle $(cat bundle-artifacts/next_version) &&
+          git reset --hard $(cat bundle-artifacts/next_version)
+      - name: Prepare home directory for code-signing
+        env:
+          CODESIGN_P12: ${{secrets.CODESIGN_P12}}
+          CODESIGN_PASS: ${{secrets.CODESIGN_PASS}}
+        if: env.CODESIGN_P12 != '' && env.CODESIGN_PASS != ''
+        run: |
+          cd home &&
+          mkdir -p .sig &&
+          echo -n "$CODESIGN_P12" | tr % '\n' | base64 -d >.sig/codesign.p12 &&
+          echo -n "$CODESIGN_PASS" >.sig/codesign.pass
+          git config --global alias.signtool '!sh "/usr/src/build-extra/signtool.sh"'
+      - name: Prepare home directory for GPG signing
+        if: env.GPGKEY != ''
+        run: |
+          echo '${{secrets.PRIVGPGKEY}}' | tr % '\n' | gpg $GPG_OPTIONS --import &&
+          info="$(gpg --list-keys --with-colons "${GPGKEY%% *}" | cut -d : -f 1,10 | sed -n '/^uid/{s|uid:||p;q}')" &&
+          git config --global user.name "${info% <*}" &&
+          git config --global user.email "<${info#*<}"
+        env:
+          GPGKEY: ${{secrets.GPGKEY}}
+      - name: Cache ${{env.MINGW_PACKAGE_PREFIX}}-git
+        id: cache-git-pkg
+        uses: actions/cache@v3
+        with:
+          path: artifacts
+          key: pkg-${{env.GIT_VERSION}}-${{env.ARCHITECTURE}}
+      - name: Build ${{env.MINGW_PACKAGE_PREFIX}}-git
+        if: steps.cache-git-pkg.outputs.cache-hit != 'true'
+        env:
+          GPGKEY: "${{secrets.GPGKEY}}"
+        run: |
+          set -x
+          BUILD_SRC=$(test x86_64 != "$ARCHITECTURE" || echo "--build-src-pkg")
+          # Make sure that there is a `/usr/bin/git` that can be used by `makepkg-mingw`
+          printf '#!/bin/sh\n\nexec '$MINGW_PREFIX'/bin/git.exe "$@"\n' >/usr/bin/git &&
+          /usr/src/build-extra/please.sh build-mingw-w64-git --only-$ARCHITECTURE $BUILD_SRC -o artifacts HEAD &&
+          cp bundle-artifacts/ver artifacts/ &&
+          if test -n "$GPGKEY"
+          then
+            for tar in artifacts/*.tar*
+            do
+              /usr/src/build-extra/gnupg-with-gpgkey.sh --detach-sign --no-armor $tar
+            done
+          fi &&
+          b=$PWD/artifacts &&
+          version=$(cat bundle-artifacts/next_version) &&
+          (cd /usr/src/MINGW-packages/mingw-w64-git &&
+          cp PKGBUILD.$version PKGBUILD &&
+          git commit -s -m "mingw-w64-git: new version ($version)" PKGBUILD &&
+          git bundle create "$b"/MINGW-packages.bundle origin/main..main)
+      - name: Publish ${{env.MINGW_PACKAGE_PREFIX}}-git
+        uses: actions/upload-artifact@v3
+        with:
+          name: pkg-${{env.ARCHITECTURE}}
+          path: artifacts
+  artifacts:
+    runs-on: windows-latest
+    needs: pkg
+    env:
+      MSYSTEM: ${{ needs.pkg.outputs.msystem }}
+      MINGW_PACKAGE_PREFIX: ${{ needs.pkg.outputs.mingw_package_prefix }}
+      SDK_REPO_ARCH: ${{ needs.pkg.outputs.sdk_repo_arch }}
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.pkg.outputs.artifact_matrix) }}
+    steps:
+      - name: Download pkg-${{env.ARCHITECTURE}}
+        uses: actions/download-artifact@v3
+        with:
+          name: pkg-${{env.ARCHITECTURE}}
+          path: pkg-${{env.ARCHITECTURE}}
+      - name: Download bundle-artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: bundle-artifacts
+          path: bundle-artifacts
+      - uses: git-for-windows/setup-git-for-windows-sdk@v1
+        with:
+          flavor: build-installers
+          architecture: ${{env.ARCHITECTURE}}
+      - name: Clone and update build-extra
+        run: |
+          d=/usr/src/build-extra &&
+          if test ! -d $d/.git
+          then
+            git clone --single-branch -b main https://github.com/git-for-windows/build-extra $d
+          else
+            git -C $d fetch https://github.com/git-for-windows/build-extra main &&
+            git -C $d switch -C main FETCH_HEAD
+          fi &&
+          git -C $d pull "$PWD"/bundle-artifacts/build-extra.bundle main
+      - name: Prepare home directory for code-signing
+        env:
+          CODESIGN_P12: ${{secrets.CODESIGN_P12}}
+          CODESIGN_PASS: ${{secrets.CODESIGN_PASS}}
+        if: (matrix.artifact.name == 'installer' || matrix.artifact.name == 'portable') && env.CODESIGN_P12 != '' && env.CODESIGN_PASS != ''
+        run: |
+          mkdir -p home/.sig &&
+          echo -n "$CODESIGN_P12" | tr % '\n' | base64 -d >home/.sig/codesign.p12 &&
+          echo -n "$CODESIGN_PASS" >home/.sig/codesign.pass &&
+          git config --global alias.signtool '!sh "/usr/src/build-extra/signtool.sh"'
+      - name: Build ${{env.ARCHITECTURE}} ${{matrix.artifact.name}}
+        run: |
+          set -x
+          eval /usr/src/build-extra/please.sh make_installers_from_mingw_w64_git --version=$(cat pkg-${{env.ARCHITECTURE}}/ver) -o artifacts --${{matrix.artifact.name}} --pkg=pkg-${{env.ARCHITECTURE}}/${{env.MINGW_PACKAGE_PREFIX}}-git-[0-9]*.tar.xz --pkg=pkg-${{env.ARCHITECTURE}}/${{env.MINGW_PACKAGE_PREFIX}}-git-doc-html-[0-9]*.tar.xz &&
+          if test portable = '${{matrix.artifact.name}}' && test -n "$(git config alias.signtool)"
+          then
+            git signtool artifacts/PortableGit-*.exe
+          fi &&
+          openssl dgst -sha256 artifacts/${{matrix.artifact.fileprefix}}-*.${{matrix.artifact.fileextension}} | sed "s/.* //" >artifacts/sha-256.txt
+      - name: Copy package-versions and pdbs
+        if: matrix.artifact.name == 'installer'
+        run: |
+          cp /usr/src/build-extra/installer/package-versions.txt artifacts/ &&
+          a=$PWD/artifacts &&
+          p=$PWD/pkg-${{env.ARCHITECTURE}} &&
+          (cd /usr/src/build-extra &&
+          mkdir -p cached-source-packages &&
+          cp "$p"/*-pdb* cached-source-packages/ &&
+          GIT_CONFIG_PARAMETERS="'windows.sdk${{env.SDK_REPO_ARCH}}.path='" ./please.sh bundle_pdbs --arch=${{env.ARCHITECTURE}} --directory="$a" installer/package-versions.txt)
+      - name: Publish ${{matrix.artifact.name}}-${{env.ARCHITECTURE}}
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{matrix.artifact.name}}-${{env.ARCHITECTURE}}
+          path: artifacts
--- a/.github/workflows/tag-git.yml
+++ b/.github/workflows/tag-git.yml
@@ -1,0 +1,73 @@
+name: tag-git
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Optionally override which branch to build'
+        required: false
+        default: main
+      repository:
+        description: 'Optionally override from where to fetch the specified ref'
+        required: false
+        default: git-for-windows/git
+
+env:
+  GPG_OPTIONS: "--batch --yes --no-tty --list-options no-show-photos --verify-options no-show-photos --pinentry-mode loopback"
+  HOME: "${{github.workspace}}\\home"
+  USERPROFILE: "${{github.workspace}}\\home"
+  REPOSITORY: "${{github.event.inputs.repository}}"
+  REF: "${{github.event.inputs.ref}}"
+  NODEJS_VERSION: 16
+
+jobs:
+  tag-git:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure user
+        run:
+          USER_NAME="${{github.actor}}" &&
+          USER_EMAIL="${{github.actor}}@users.noreply.github.com" &&
+          mkdir "$HOME" &&
+          git config --global user.name "$USER_NAME" &&
+          git config --global user.email "$USER_EMAIL" &&
+          echo "PACKAGER=$USER_NAME <$USER_EMAIL>" >> $GITHUB_ENV
+      - name: Clone build-extra
+        run: git clone --single-branch -b main https://github.com/git-for-windows/build-extra $RUNNER_TEMP/build-extra
+      - name: Prepare home directory for GPG signing
+        if: env.GPGKEY != ''
+        run: |
+          echo '${{secrets.PRIVGPGKEY}}' | tr % '\n' | gpg $GPG_OPTIONS --import &&
+          mkdir -p home &&
+          git config --global gpg.program "$RUNNER_TEMP/build-extra/gnupg-with-gpgkey.sh" &&
+          info="$(gpg --list-keys --with-colons "${GPGKEY%% *}" | cut -d : -f 1,10 | sed -n '/^uid/{s|uid:||p;q}')" &&
+          git config --global user.name "${info% <*}" &&
+          git config --global user.email "<${info#*<}"
+        env:
+          GPGKEY: ${{secrets.GPGKEY}}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{env.NODEJS_VERSION}}
+      - name: Generate bundle artifacts
+        env:
+          GPGKEY: ${{secrets.GPGKEY}}
+        run: |
+          mkdir -p bundle-artifacts &&
+          git -c init.defaultBranch=main init --bare &&
+          git remote add -f origin https://github.com/git-for-windows/git &&
+          git fetch "https://github.com/$REPOSITORY" "$REF:$REF" &&
+          tag_name="$(git describe --match 'v[0-9]*' FETCH_HEAD)" &&
+          echo "prerelease-${tag_name#v}" >bundle-artifacts/ver &&
+          echo "${tag_name#v}" >bundle-artifacts/display_version &&
+          echo "$tag_name" >bundle-artifacts/next_version &&
+          git tag $(test -z "$GPGKEY" || echo " -s") -m "Snapshot build" "$tag_name" FETCH_HEAD &&
+          git bundle create bundle-artifacts/git.bundle origin/main.."$tag_name" &&
+          release_note=$(git show -s  --pretty='tformat:%h (%s, %ad)' --date=short FETCH_HEAD) &&
+          cd $RUNNER_TEMP/build-extra &&
+          node ./add-release-note.js --commit feature "Snapshot of $release_note" &&
+          git bundle create "$GITHUB_WORKSPACE/bundle-artifacts/build-extra.bundle" origin/main..main
+      - name: 'Publish Pipeline Artifact: bundle-artifacts'
+        uses: actions/upload-artifact@v3
+        with:
+          name: bundle-artifacts
+          path: ${{ github.workspace }}/bundle-artifacts
--- a/create-artifacts-matrix.js
+++ b/create-artifacts-matrix.js
@@ -1,0 +1,53 @@
+module.exports = (core, artifactsString, architecture) => {
+    const artifacts = artifactsString.split(' ')
+
+    if (artifacts.length < 1) {
+        core.setFailed('No artifacts provided. Provide in a space-separated string, e.g. "installer portable"')
+        return
+    }
+
+    const validArtifacts = [
+        {
+            name: 'installer',
+            filePrefix: 'Git',
+            fileExtension: 'exe'
+        },
+        {
+            name: 'portable',
+            filePrefix: 'PortableGit',
+            fileExtension: 'exe'
+        },
+        {
+            name: 'archive',
+            filePrefix: 'Git',
+            fileExtension: 'tar.bz2'
+        },
+        {
+            name: 'mingit',
+            filePrefix: 'MinGit',
+            fileExtension: 'zip'
+        },
+        architecture !== 'aarch64' && {
+            name: 'mingit-busybox',
+            filePrefix: 'MinGit',
+            fileExtension: 'zip'
+        }
+    ]
+    
+    const artifactsToBuild = []
+    
+    for (const artifact of artifacts) {
+        const artifactObject = validArtifacts.find(a => a.name === artifact)
+        if (!artifactObject) {
+            core.setFailed(`${artifact} is not a valid artifact for ${architecture}`)
+            return
+        }
+    
+        artifactsToBuild.push(artifactObject)
+    }
+
+    const output = {artifact: artifactsToBuild}
+
+    core.info(`Will be using the following matrix: ${JSON.stringify(output)}`)
+    core.setOutput('matrix', output)
+}
--- a/get-bundle-artifacts-artifact.js
+++ b/get-bundle-artifacts-artifact.js
@@ -1,0 +1,40 @@
+module.exports = async (github, context, core, workflowId) => {
+    const { data } = workflowId
+        ? await github.rest.actions.listWorkflowRunArtifacts({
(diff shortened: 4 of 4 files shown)


.github/workflows/git-artifacts.yml
```diff
@@ -0,0 +1,313 @@
...
+        if: steps.cache-git-pkg.outputs.cache-hit != 'true'
+        env:
+          GPGKEY: "${{secrets.GPGKEY}}"
+        run: |
+          set -x
+          # Make sure that there is a `/usr/bin/git` that can be used by `makepkg-mingw`
+          printf '#!/bin/sh\n\nexec $MINGW_PREFIX/bin/git.exe "$@"\n' >/usr/bin/git &&
+          # Restrict `PATH` to MSYS2 and to Visual Studio (to let `cv2pdb` find the relevant DLLs)
+          PATH="$MINGW_PREFIX/bin:/usr/bin:/c/Program Files/Microsoft Visual Studio/2022/Enterprise/Common7/IDE/:/C/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin$MSVC_FOLDER_SUFFIX:/C/Windows/system32"
+          type -p mspdb140.dll || exit 1
```
dennisameling, line 211: Is this still needed in 2023? If yes, maybe `vswhere` will be a bit more robust as it prevents us from having to hardcode VS installation paths. WDYT?

.github/workflows/git-artifacts.yml
```diff
@@ -0,0 +1,313 @@
...
+      GIT_VERSION: ${{ needs.bundle-artifacts.outputs.git_version }}
+    steps:
+      - name: Configure user
+        run:
+          USER_NAME="${{github.actor}}" &&
+          USER_EMAIL="${{github.actor}}@users.noreply.github.com" &&
+          mkdir "$HOME" &&
+          git config --global user.name "$USER_NAME" &&
+          git config --global user.email "$USER_EMAIL" &&
+          echo "PACKAGER=$USER_NAME <$USER_EMAIL>" >> $GITHUB_ENV
```
dennisameling, line 147: We already did this part in the `bundle-artifacts` job. I'd be in favor of combining the `bundle-artifacts` and `pkg` jobs to prevent us from having to do this duplicate stuff. WDYT?

.github/workflows/git-artifacts.yml
```diff
@@ -0,0 +1,313 @@
...
+        description: 'Optionally override from where to fetch the specified ref'
+        required: false
+        default: git-for-windows/git
+      architecture:
+        type: choice
+        description: 'Architecture to build'
+        required: true
+        options: 
+        - x86_64
+        - i686
```
dennisameling, line 23: I chose to support only a single architecture at a time. Supporting multi-arch builds would make the workflow slightly more complex, but it's definitely doable. WDYT?

I'll add `arm64` in a follow-up PR because `please.sh build-mingw-w64-git` doesn't support that architecture yet. https://github.com/git-for-windows/build-extra/pull/452 will add that support.

.github/workflows/git-artifacts.yml
dscho: This is _not_ needed. We know that Visual Studio is available on the hosted runners and that `cv2pdb` has no trouble finding them thanks to https://github.com/rainers/cv2pdb/pull/66 (and we do not need `cv2pdb` let alone `mspdb140.dll` in clang-aarch64 builds).

Thank you for pointing that out!

.github/workflows/git-artifacts.yml
dscho: It would probably make a lot of sense to combine them, indeed. It's not as if we're building the packages in different architectures, we use `git-sdk-64` to build both i686 and x86_64 builds.

.github/workflows/git-artifacts.yml
dscho: > Supporting multi-arch builds would make the workflow slightly more complex, but it's definitely doable.

I would love to have a multi-arch build, much [like we do for the MINGW-packages](https://github.com/git-for-windows/git-for-windows-automation/blob/e9b20e2697a348fd7c078fbe6a8f3c05e315a734/.github/workflows/build-and-deploy.yml#L254).

.github/workflows/git-artifacts.yml
```diff
@@ -0,0 +1,313 @@
...
+            *)
+              echo "Unhandled architecture: $ARCHITECTURE"
+              exit 1
+              ;;
+          esac
+          echo "MSYSTEM=$MSYSTEM" >> $GITHUB_ENV
+          echo "MSYSTEM=$MSYSTEM" >> $GITHUB_OUTPUT
+          echo "MINGW_PREFIX=$MINGW_PREFIX" >> $GITHUB_ENV
+          echo "MINGW_PREFIX=$MINGW_PREFIX" >> $GITHUB_OUTPUT
+          echo "BITNESS=$BITNESS" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
dscho: The biggest trick with arm64 will be to avoid creating a new tag, and instead re-using the one from the corresponding i686/x86_64 build...

.github/workflows/git-artifacts.yml
```diff
@@ -0,0 +1,314 @@
+name: git-artifacts
+
+on:
+  workflow_dispatch:
+    inputs:
+      build_only:
```
dscho, line 6: `build_only` sounds like a Boolean. This is totally my fault. But maybe we can fix it "while at it", maybe `artifacts`

.github/workflows/git-artifacts.yml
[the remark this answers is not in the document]
dscho: Maybe we should use the opportunity to switch things up a bit more and decouple the build from creating the tag? Dunno.

An alternative would be to do what `microsoft/git` does, which creates releases triggered by tags that are pushed to the repository.

In Git for Windows, it is done differently, for two reasons:
- people might fetch the tag prematurely, i.e. before the build succeeded, which means that the build might fail and require more work and the tag would then have to be updated. That can happen e.g. when anything on [the "pre-flight checklist"](https://github.com/git-for-windows/build-extra/blob/main/installer/checklist.txt) fails. It did happen in the past.
- In Git for Windows, we have to deal with embargoed releases (like v2.39.1, last week), where we have to create the artifacts a couple of days in advance, including the tag, and nothing about that must leak to the public before the actual release is published.

So maybe a better alternative would be a separate GitHub workflow that merely creates the tag and the release notes (essentially the `bundle-artifacts` of, say, [the v2.39.0(2) build](https://dev.azure.com/git-for-windows/git/_build/results?buildId=111287&view=artifacts&pathAsName=false&type=publishedArtifacts)).

The `git-artifacts` workflow would then require a reference to said build so that it can download the artifact and use it.

And of course we will eventually also need to port [the Azure Release Pipeline](https://dev.azure.com/git-for-windows/git/_release?_a=releases&view=mine&definitionId=1) that uploads/publishes the artifacts and sends the announcement email. But we can deal with that when the first two workflows are in place.

.github/workflows/git-artifacts.yml
dscho: On second thought, this presents a fine opportunity to clean things up and to use `git-sdk-32` to build the `mingw-w64-i686-git` packages...

dscho: Yep, I think you're right, @dennisameling, we should support only a single architecture at a time. That will make things so much easier.

.github/workflows/git-artifacts.yml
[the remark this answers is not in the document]
dennisameling: @dscho I've now split up the workflow into `bundle-artifacts.yml` and `git-artifacts.yml`. If no workflow run id of `bundle-artifacts.yml` is provided, it simply takes the latest available one. WDYT?

- Example run of `bundle-artifacts.yml`: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4036805037
- Example run of `git-artifacts.yml` without `bundle_artifacts_workflow_run_id`: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4037281758
- Example run of `git-artifacts.yml` with `bundle_artifacts_workflow_run_id` of `4036805037`: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4037281503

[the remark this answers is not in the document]
dennisameling: @dscho thanks to your changes [here](https://github.com/git-for-windows/git-sdk-32/pull/22) and [here](https://github.com/git-for-windows/build-extra/pull/473), the `i686` run is now progressing a lot further: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4047585094/jobs/6961778017

However, it's still failing because it was invoked with `--build-src-pkg` [which sets `MINGW_ARCH=mingw64`](https://github.com/git-for-windows/build-extra/blob/9544308613847e67a6863e32d080d701581b5a3a/please.sh#L3695-L3698). That obviously doesn't work in the `git-sdk-32`:

```
==> Finished making: mingw-w64-git 2.39.1.windows.1.529.g10a93f8bfd-1 (Mon Jan 30 20:07:31 2023)
=> WARNING: You don't have the required toolchain installed for mingw64.
=> WARNING: To install it run: 'pacman -S mingw-w64-x86_64-toolchain'
==> Making package: mingw-w64-git 2.39.1.windows.1.529.g10a93f8bfd-1 (Mon Jan 30 20:13:04 2023)
==> Retrieving sources...
/usr/bin/git: line 3: /mingw64/bin/git.exe: No such file or directory
==> ERROR: /usr/src/MINGW-packages/mingw-w64-git/git is not a clone of https://github.com/git-for-windows/git.git
    Aborting...
Could not build mingw-w64-git
```

What's your preferred approach for this? Only build the source package on `x86_64`? Happy to provide a PR but would first like to know which way to go. Thanks!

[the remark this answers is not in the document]
dscho: > What's your preferred approach for this? Only build the source package on `x86_64`?

I think that's the best course of action, yes.

.github/workflows/bundle-artifacts.yml
```diff
@@ -0,0 +1,83 @@
...
+  REPOSITORY: "${{github.event.inputs.repository}}"
+  REF: "${{github.event.inputs.ref}}"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  bundle-artifacts:
+    runs-on: windows-latest
```
dscho, line 28: I am _so_ itching to move this to `ubuntu-latest`, to avoid paying the startup/runtime cost of the slower Windows agents... Maybe it works with just this change?

dennisameling: That actually makes a ton of sense indeed. I'm not seeing anything in these scripts that `ubuntu-latest` wouldn't be able to do. We'll need to get rid of the `setup-git-for-windows-sdk` step and some other specific logic, but nothing major there. Let me try. Thanks for the suggestion!

dennisameling: Ok, it's [almost working](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4107209885/jobs/7086405513). Getting this error:

```
error: gpg failed to sign the data
error: unable to sign the tag
```

Not a very helpful error message. Have you run into this before?

dscho: This should help: https://github.com/git-for-windows/build-extra/pull/476

dscho:
Could we also rename the workflow? `bundle-artifacts` was kind of okay for a step in the `git-artifacts` workflow (even if I never was really happy with the non-descriptive name), but if we extract this into its stand-alone workflow, I think we'll want a better name.

How about `tag-git`?

.github/workflows/bundle-artifacts.yml
dennisameling: That [helped a lot](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4113180504), but we're not there yet. The script tries to `cd` into `/usr/src/build-extra`, but that doesn't work as I cloned it into `$RUNNER_TEMP/build-extra` instead. When I let CI clone into `/usr/src/build-extra` in one of my first attempts yesterday, I got a [Permission Denied](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4107164239/jobs/7086298667#step:3:19) error. I still think `$RUNNER_TEMP/build-extra` would be a good place for this as it's guaranteed that we have write permissions in that folder. We'd need to update `please.sh` a bit though. WDYT?

dscho: How about using https://github.com/git-for-windows/build-extra/blob/main/add-release-note.js instead of `please.sh mention`? It's the long-term plan to retire the latter in favor of the former, anyway.

dennisameling: Neat!! Looks like [that perfectly](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4115147857/jobs/7103465335). Took a few attempts to get right, sorry for all the pushes in this PR 😅

dennisameling:
@dscho, very excited to mention that this is now working on all three architectures with their own, dedicated SDKs 🚀🎉😄 

- `tag-git`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4115168424) (on `ubuntu-latest`!!!) ✅ 
- `git-artifacts x86_64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4116285102) ✅ 
- `git-artifacts i686`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4116214355) ✅ 
- `git-artifacts aarch64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4116530926/jobs/7107199395) ✅ (except for `installer` and `mingit-busybox` for which I'll need to provide fixes in `build-extra`)

One issue I encountered is that arm64 runners [cannot be deployed in parallel](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4116757131/attempts/1) due to duplicate deployment names (`azure-arm-template`). [Let's set the deployment name explicitly](https://github.com/git-for-windows/git-for-windows-automation/pull/26), which should hopefully prevent these issues.

Once the first "official" GfW arm64 version is released, let's also update the Azure VM [post-deployment script](https://github.com/git-for-windows/git-for-windows-automation/blob/c701662ff6838bea2ff4fccccf73df3b2af95605/azure-self-hosted-runners/post-deployment-script.ps1#L48) so that it downloads the `arm64` version instead of `amd64`. This will benefit some CI workflows (or parts of those) that don't use the `git-sdk-arm64` but rely on the system-installed Git.

dennisameling:
> (except for `installer` and `mingit-busybox` for which I'll need to provide fixes in build-extra)

Ok, let's not support `mingit-busybox` on arm64 for now. It's based on the package in [`MINGW-packages/mingw-w64-busybox`](https://github.com/git-for-windows/MINGW-packages/tree/main/mingw-w64-busybox) which in turn builds [`git-for-windows/w32-busybox`](https://github.com/git-for-windows/busybox-w32). That has configs for [`mingw32`](https://github.com/git-for-windows/busybox-w32/blob/main/configs/mingw32_defconfig) and [`mingw64`](https://github.com/git-for-windows/busybox-w32/blob/main/configs/mingw64_defconfig), but not seeing anything `clang`-related there. [I tried adding `clangarm64_defconfig`](https://github.com/dennisameling/busybox-w32/tree/windows-arm64), but it looks like that'll need quite a bit more fixes. Someone should probably first try to build `clang64` before even considering `clangarm64`. Even the author of the upstream repo [hasn't been able to create a _working_ build using the Clang toolchain yet](https://github.com/rmyorston/busybox-w32/issues/239#issuecomment-1296222961). Shall I just add some logic to completely skip the `mingit-busybox` artifact if we're on arm64? I don't have the bandwidth now to dive into this further.

Regarding the `installer` artifact: the artifact itself can be built already, but the part that fails is `./please.sh bundle_pdbs --arch=aarch64`. That method [currently only supports `bitness`](https://github.com/git-for-windows/build-extra/blob/9cf3a5a7273a9390f2aee523ef834c3f7f6f7011/please.sh#L2811-L2946) so will need a bit of refactoring to support arm64 as well.

dscho:
> Shall I just add some logic to completely skip the `mingit-busybox` artifact if we're on arm64?

I think that's the most sensible course of action.

> Regarding the `installer` artifact: the artifact itself can be built already, but the part that fails is `./please.sh bundle_pdbs --arch=aarch64`. That method [currently only supports `i686`/`x86_64`](https://github.com/git-for-windows/build-extra/blob/9cf3a5a7273a9390f2aee523ef834c3f7f6f7011/please.sh#L2811-L2946) so will need a bit of refactoring to support `aarch64` as well.

Right. It looks as if that might be the last big obstacle, though, amirite?

dennisameling:
> Right. It looks as if that might be the last big obstacle, though, amirite?

Yes!!! 😍 Will provide a PR later today so we can push all of this over the finish line. So exciting!

dennisameling:
> Yes!!! 😍 Will provide a PR later today so we can push all of this over the finish line. So exciting!

[Here we go](https://github.com/git-for-windows/build-extra/pull/478)!

dennisameling:
@dscho here we go!!!! 🚀 🔥 🎉 

- `git-artifacts x86_64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144314732) ✅ 
- `git-artifacts i686`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144315321) ✅ 
- `git-artifacts aarch64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144559603) ✅ 

Small note: while the `arm64` runners are created in parallel now, your Azure subscription [seems to hit a quota of CPU cores](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144779320/jobs/7168260908) 😅 I guess that just means that too many VMs are active at the same time. The limit is 20, so that means that two 8-core VMs would be the maximum already. We could use the 4-core machines instead so there can be 5 active VMs at the same time. WDYT?

dscho:
> here we go!!!! 🚀 🔥 🎉
> 
>     * `git-artifacts x86_64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144314732) ✅
> 
>     * `git-artifacts i686`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144315321) ✅
> 
>     * `git-artifacts aarch64`: [successful run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144559603) ✅

Yay! Time to merge this PR, then?

> your Azure subscription [seems to hit a quota of CPU cores](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144779320/jobs/7168260908) 😅 I guess that just means that too many VMs are active at the same time. The limit is 20, so that means that two 8-core VMs would be the maximum already. We could use the 4-core machines instead so there can be 5 active VMs at the same time.

That sounds very sensible to me.

dennisameling:
> Yay! Time to merge this PR, then?

Sounds good! 🚀 

> That sounds very sensible to me.

Okay. [Here we go](https://github.com/git-for-windows/git-for-windows-automation/pull/30)!

dscho:
Hooray!

dennisameling:
Yay! Anything I can help with still to get the arm64 binaries in one of the upcoming releases?

Two minor things from my side:
- Let's [remove the `git-artifacts` pipeline from the main repo](https://github.com/git-for-windows/git/pull/4282) as it's no longer needed there.
- I have some pending documentation updates regarding the earlier `git-extra` changes. Looks like it's not possible to show that in some UI, but you can clone `https://github.com/git-for-windows/git.wiki.git` and look for the `update-git-extra-docs` branch. 😊

dscho:
> * I have some pending documentation updates regarding the earlier `git-extra` changes. Looks like it's not possible to show that in some UI, but you can clone `https://github.com/git-for-windows/git.wiki.git` and look for the `update-git-extra-docs` branch. 😊

Merged & pushed 😃

dennisameling:
@dscho how is the `tag-git` pipeline supposed to work for existing tags (e.g. v2.39.2.windows.1)? I tried both the commit ref and the tag on that pipeline, but both failed ([here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4177561690/jobs/7248895800) and [here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4183940000/jobs/7248939363)). I can't kick off a `git-artifacts` pipeline for `aarch64` before having run `bundle-artifacts`. It should probably just take the existing tag and create the `bundle-artifacts` artifact, right?

dscho:
@dennisameling since building Git for Windows/ARM64 is probably not going to be a recurrent thing, now that `tag-git` and `git-artifacts` are working, how about simply branching off, say, `git-2.39.2-aarch64`, and modifying the workflow as a one-off? I.e.

- replace the `workflow_dispatch:` section by a single `push:`
- edit the global `env` section to hard-code the values that would have been `workflow_dispatch` inputs as well as the `runs-on:` directives (search for `github.event.inputs`)
- remove the `Get bundle-artifacts download URL` and `Download bundle-artifacts zip` steps and replace it with a step that creates a `bundle-artifacts/` directory containing:
  - `build-extra.bundle` (probably as simple as `git --git-dir=/usr/src/build-extra/.git bundle create bundle-artifacts/build-extra.bundle main^..main`)
  - `git.bundle` (_maybe_ as simple as `git --git-dir=/usr/src/git bundle create bundle-artifacts/git.bundle origin/main^..v2.39.2.windows.1`, but maybe there is no clone yet and you might have to jump through hoops like `git init --bare .tmp.git && git ---git-dir .tmp.git fetch --filter=blob:none --depth=1 refs/tags/v2.39.2.windows.1:refs/tags/v2.39.2.windows.1`)
  - `display_version`, `next_version` and `ver` (I _think_ they contain the same text, you can call `echo "2.39.2.windows.1" >bundle-artifacts/display_version`)

Then push that branch to `git-for-windows/git-for-windows-automation` and let it build the artifacts. Does that sound like a good plan to you?

dscho:
@dennisameling after thinking about this, I believe that it makes most sense to _add_ that functionality to the workflow instead of keeping it only in a topic branch (and then essentially dropping it via `-s ours`). I.e. I want to introduce a new `input` that takes the name of an existing tag and another one for the `build-extra` revision to use. I'll get to it immediately and report back.

dscho:
> I'll get to it immediately and report back.

[Here](https://github.com/git-for-windows/git-for-windows-automation/pull/34) is the PR, and [here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4188801813) is the workflow run that I just started to build the Git for Windows v2.39.2 artifacts targeting ARM64.

dscho:
It's hanging again. Let me document better what I did. First of all, I RDP'ed into the machine, then opened a Git Bash. In that Bash, I looked for the process that would be hanging by first calling `wmic process where 'CommandLine like "%git%"' get ProcessId,ParentProcessId,CommandLine`, seeing that it's probably the signtool script, and then calling:

```console
$ wmic process where 'CommandLine like "%sign%"' get ProcessId,ParentProcessId,CommandLine
CommandLine                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            ParentProcessId  ProcessId
D:\git-sdk-arm64-build-installers\usr\bin\make.exe -f ../mingw-w64-git.mak sign-executables                                                                                                                                                                                                                                                                                                                                                                                                                                                            4408             6100
D:\git-sdk-arm64-build-installers\usr\bin\sh.exe -c "eval git --git-dir=D:/a/_work/git-for-windows-automation/git-for-windows-automation/.git signtool headless-git.exe git-daemon.exe git-http-backend.exe git-imap-send.exe git-sh-i18n--envsubst.exe git-shell.exe git-http-fetch.exe git-http-push.exe git-remote-http.exe git-remote-https.exe git-remote-ftp.exe git-remote-ftps.exe \    contrib/credential/wincred/git-credential-wincred.exe git.exe \ cmd/git{,-gui,k}.exe compat-bash.exe git-{bash,cmd,wrapper}.exe"                          1872             2684
D:\git-sdk-arm64-build-installers\usr\bin\sh.exe -c "eval git --git-dir=D:/a/_work/git-for-windows-automation/git-for-windows-automation/.git signtool headless-git.exe git-daemon.exe git-http-backend.exe git-imap-send.exe git-sh-i18n--envsubst.exe git-shell.exe git-http-fetch.exe git-http-push.exe git-remote-http.exe git-remote-https.exe git-remote-ftp.exe git-remote-ftps.exe \    contrib/credential/wincred/git-credential-wincred.exe git.exe \ cmd/git{,-gui,k}.exe compat-bash.exe git-{bash,cmd,wrapper}.exe"                          2684             5676
D:\git-sdk-arm64-build-installers\clangarm64\bin\git.exe --git-dir=D:/a/_work/git-for-windows-automation/git-for-windows-automation/.git signtool headless-git.exe git-daemon.exe git-http-backend.exe git-imap-send.exe git-sh-i18n--envsubst.exe git-shell.exe git-http-fetch.exe git-http-push.exe git-remote-http.exe git-remote-https.exe git-remote-ftp.exe git-remote-ftps.exe contrib/credential/wincred/git-credential-wincred.exe git.exe cmd/git.exe cmd/git-gui.exe cmd/gitk.exe compat-bash.exe git-bash.exe git-cmd.exe git-wrapper.exe  5676             5200
sh -c "sh \"/usr/src/build-extra/signtool.sh\" \"$@\"" "sh \"/usr/src/build-extra/signtool.sh\"" headless-git.exe git-daemon.exe git-http-backend.exe git-imap-send.exe git-sh-i18n--envsubst.exe git-shell.exe git-http-fetch.exe git-http-push.exe git-remote-http.exe git-remote-https.exe git-remote-ftp.exe git-remote-ftps.exe contrib/credential/wincred/git-credential-wincred.exe git.exe cmd/git.exe cmd/git-gui.exe cmd/gitk.exe compat-bash.exe git-bash.exe git-cmd.exe git-wrapper.exe                                                   5200             5024
C:\Windows\System32\Wbem\wmic.exe process where "CommandLine like \"%sign%\"" get ProcessId,ParentProcessId,CommandLine                                                                                                                                                                                                                                                                                                                                                                                                                                10112            10016
```

So indeed, there were quite a few, and the process with ID 5024 seemed to be the bottom-most one. To verify, I called:

```console
$ wmic process where 'ParentProcessId=5024' get ProcessId,ParentProcessId,CommandLine
No Instance(s) Available.
```

Okay. Now on to debug. For that, I need `gdb.exe`. So I cloned the entire SDK using the command-line `git -C /d/ clone --depth=1 https://github.com/git-for-windows/git-sdk-arm64` in the Git Bash. After that, I opened a Git SDK Bash by pressing the Windows button and then typing `D:\git-sdk-arm64\git-bash.exe` <kbd>Enter</kbd>.

Unfortunately, that does not work because of the OpenSSL situation where Git for Windows stays with v1.1.1* as long as possible while MSYS2 already switched to v3.*:

```console
$ gdb.exe
D:/git-sdk-arm64/usr/bin/gdb.exe: error while loading shared libraries: msys-python3.11.dll: cannot open shared object file: No such file or directory
```

So I downloaded MSYS2's `python` package via `curl -LO https://repo.msys2.org/msys/x86_64/python-3.11.2-1-x86_64.pkg.tar.zst` and then installed it via `pacman -U ./python-3.11.2-1-x86_64.pkg.tar.zst`. This let me start `gdb.exe bash.exe 5024` (that stanza asks `gdb` to load the `gdb.exe` binary and attach to the process with ID 5024).

In that GDB session, I then tried to see the current threads:

```
(gdb) info thread
  Id   Target Id          Frame
  1    Thread 5024.0x10fc error return ../../gdb-11.1/gdb/windows-nat.c:609 was 31
0x0000000000000000 in ?? ()
  2    Thread 5024.0x1978 0x00007fff30ad6a64 in ?? ()
  3    Thread 5024.0x18a8 0x00007fff30ad69c4 in ?? ()
  4    Thread 5024.0x6a8  error return ../../gdb-11.1/gdb/windows-nat.c:609 was 31
0x0000000000000000 in ?? ()
  5    Thread 5024.0x1468 0x00007fff30ad69c4 in ?? ()
* 6    Thread 5024.0x2058 0x00007fff30ae2374 in ?? ()
```

Not very informative, so let's try to see the current thread's stacktrace:

```
(gdb) bt
#0  0x00007fff30ae2374 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

That's it.

My next attempt involved downloading [Sysinternals' Process Explorer](https://learn.microsoft.com/en-us/sysinternals/downloads/process-explorer). That gave me more information. Apparently there is one thread with this stack trace:

```
ntoskrnl.exe!ZwUpdateWnfStateData+0xaf60
ntoskrnl.exe!KeUpdateThreadTag+0xdc98
ntoskrnl.exe!KiDispatchInterrupt+0x7298
ntoskrnl.exe!KeWaitForMutexObject+0x440
ntoskrnl.exe!KeUpdateThreadTag+0xa754
ntoskrnl.exe!KiDeliverApc+0x18c
ntoskrnl.exe!KeUpdateThreadTag+0xdd1c
ntoskrnl.exe!KiDispatchInterrupt+0x7298
ntoskrnl.exe!KeWaitForMutexObject+0x440
ntoskrnl.exe!ObWaitForSingleObject+0xb4
ntoskrnl.exe!NtWaitForSingleObject+0x34
ntoskrnl.exe!ZwUpdateWnfStateData+0xadf0
ntoskrnl.exe!ZwUpdateWnfStateData+0xa9e8
!ZwWaitLowEventPair+0x158284
!ZwWaitLowEventPair+0x19d190
!ZwWaitLowEventPair+0x1950a8
!ZwWaitLowEventPair+0x1a9ad0
!ZwWaitLowEventPair+0x290f80
!ZwWaitLowEventPair+0x29ff14
!_assert+0x41c6
```

So that _looks_ as if it called `_assert()`. And another one with this stack trace:

```
ntoskrnl.exe!ZwUpdateWnfStateData+0xaf60
ntoskrnl.exe!KeUpdateThreadTag+0xdc98
ntoskrnl.exe!KiDispatchInterrupt+0x7298
ntoskrnl.exe!KeWaitForMutexObject+0x440
ntoskrnl.exe!FsRtlNotifyVolumeEventEx+0x4470
ntoskrnl.exe!NtQueryVolumeInformationFile+0xfe0
ntoskrnl.exe!NtReadFile+0xb0
ntoskrnl.exe!ZwUpdateWnfStateData+0xadf0
ntoskrnl.exe!ZwUpdateWnfStateData+0xa9e8
!ZwWaitLowEventPair+0x158324
!ZwWaitLowEventPair+0x178bc8
!PackageSidFromProductId+0x2318c0
!PackageSidFromProductId+0x26a360
!sigfillset+0x2d61
```

This looks as if it was inside the `sigfillset()` function. But that does not make _too_ much sense because [that function does not call any other function](https://github.com/git-for-windows/msys2-runtime/blob/8a2855c182ed7117bae3c4eb8a17eb8f26b27891/winsup/cygwin/signal.cc#L538-L543).

The other threads' stack trace can either not be accessed or look like they're completely outside of Bash's and MSYS2 runtime's source code:

```
ntoskrnl.exe!ZwUpdateWnfStateData+0xaf60
ntoskrnl.exe!KeUpdateThreadTag+0xdc98
ntoskrnl.exe!KiDispatchInterrupt+0x7298
ntoskrnl.exe!KeWaitForMutexObject+0x440
ntoskrnl.exe!KeUpdateThreadTag+0xa754
ntoskrnl.exe!KiDeliverApc+0x18c
ntoskrnl.exe!KeUpdateThreadTag+0xdd1c
ntoskrnl.exe!KiDispatchInterrupt+0x7298
ntoskrnl.exe!KeDelayExecutionThread+0x614
ntoskrnl.exe!PsWow64IsMachineSupported+0x2d04
ntoskrnl.exe!ZwUpdateWnfStateData+0xadf0
ntoskrnl.exe!ZwUpdateWnfStateData+0xa9e8
!ZwWaitLowEventPair+0x1616f4
!ZwWaitLowEventPair+0x1eea48
!ZwWaitLowEventPair+0x1edea8
!ZwWaitLowEventPair+0x1ebe40
!ZwWaitLowEventPair+0x18495c
!ZwWaitLowEventPair+0x187a54
!ZwWaitLowEventPair+0x1ac658
!ZwWaitLowEventPair+0x1abc70
!ZwWaitLowEventPair+0x1ad458
!uaw_wcsrchr+0x81c70
!ZwWaitLowEventPair+0x1a9bc8
```

I'm out of ideas what to investigate on that side, so I downloaded [Sysinternals' `handle` utility](https://learn.microsoft.com/en-us/sysinternals/downloads/handle) to see what handles are open in that process, but that also fails to shed any light into the issue:

```console
$ ./handle64a.exe -p 5024

Nthandle v5.0 - Handle viewer
Copyright (C) 1997-2022 Mark Russinovich
Sysinternals - www.sysinternals.com

  180: Section       \BaseNamedObjects\msys-2.0S5-c2a8dd8be8845dc5\shared.5
  184: Section       \BaseNamedObjects\msys-2.0S5-c2a8dd8be8845dc5\S-1-5-20.1
  1DC: File  (RWD)   D:\git-sdk-arm64-build-installers\usr\src\MINGW-packages\mingw-w64-git\src\git
  1EC: Section       \BaseNamedObjects\msys-2.0S5-c2a8dd8be8845dc5\cygpid.10728
  2C0: Section       \BaseNamedObjects\msys-2.0S5-c2a8dd8be8845dc5\cygpid.10728
```

So for now I'll just call it yet another unexplained hang and try to let the workflow run continue, at least, by force-stopping that process: `wmic process where ProcessID=5024 delete`.

And sure enough, the workflow run [continues](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4192755213/jobs/7268751365#step:16:1838)...

dscho:
Darn, darn, darn. _Again_ I had to RDP into the runner and force-stop [the `signtool` process](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4195194569/jobs/7274388607#step:16:1835) in order to let it complete (toggle the timestamps to see the 24 minute gap where there should be less than 10 seconds).

This is not good. In this shape, we cannot publish a Git for Windows/ARM64. If it hangs for us so often, it will hang for plenty of users, too.

dscho:
And it's hanging [again](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4195194569/jobs/7275828840#step:7:38865), and also [here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4195194569/jobs/7275829074#step:7:19590). I worked around this by RDP'ing into the machines and running essentially:

```shell
candidates= &&
for pid in $(
    wmic process where 'ExecutablePath like "%arm64-build-installers%"' get ProcessId |
    sed 1d
)
do
    test "z${pid#[0-9]}" != "z$pid" || continue
    case "$(wmic process where parentprocessid=$pid get processid 2>&1 >/dev/null)" in
     "No Instance(s) Available"*)
         candidates="$candidates $pid"
         ;;
    esac
done &&
wmic process where processid=${candidates##* } delete
```

dscho:
I reached out to `#cygwin-devel` on libera.chat, and @lazka pointed me to https://github.com/msys2/msys2-autobuild/issues/62, noting:

> it only happening for those two things and not in say autotools/configure/make seems to suggest it's not a general problem
> (those two things being gpgme signature verification, and install-info)

And @jturney said:

> "Not a known issue, but now we know about it.." 😉

lazka:
@jeremyd2019 tried to debug our hangs at one point

jeremyd2019:
yeah, I couldn't get anywhere though, none of the debuggers I tried to attach were able to get the thread context for the 'main' thread.  there was also another thread running which seemed to be for cygwin signal handling, which seemed to be happily waiting.  I gave up trying to debug it and instead added workarounds to avoid situations where it seemed to happen.  Now that somebody else is reproducing it, maybe it can be debugged.

Regarding your `_assert` and `sigfillset` observations, notice the offsets from the symbols.  It seems unlikely that those are actually the functions involved, just the nearest symbols it could find.

dscho:
> Regarding your `_assert` and `sigfillset` observations, notice the offsets from the symbols. It seems unlikely that those are actually the functions involved, just the nearest symbols it could find.

Right, I feared so much.

> Now that somebody else is reproducing it, maybe it can be debugged.

I fear that we'll have to get a much more reliable reproducer first.

_Maybe_ the following will reproduce the hang reliably enough: calling the MINGW executable `osslsigncode.exe` from the MSYS2 Bash script [`signtool.sh`](https://github.com/git-for-windows/build-extra/blob/main/signtool.sh), which in turn is called as a Git alias from the _MINGW_ executable `git.exe` that is called from either an MSYS2 Bash script or MSYS2 `make.exe`.

At least that is what hung reliable for me, at the end of the inner-most shell script just after all the code-signing was done.

Alovchin91:
Since you are RDP'ing into the machine anyway, could you maybe collect a full memory dump of a hanging process using [procdump](https://learn.microsoft.com/en-us/sysinternals/downloads/procdump)? Maybe that could shed some more light on what it's waiting for 🤔

Alovchin91:
As another (test) option, maybe use an x64 osslsigncode.exe temporarily to see if that resolves the issue?

Alovchin91:
> ... calling the MINGW executable `osslsigncode.exe` from the MSYS2 Bash script [`signtool.sh`](https://github.com/git-for-windows/build-extra/blob/main/signtool.sh), which in turn is called as a Git alias from the _MINGW_ executable `git.exe` ...

Hmm, so `git.exe` calls `signtool.sh`, which signs `git.exe`, produces `git.signed.exe` (according to the script), and then moves it in place of the original?

dscho:
@Alovchin91 it's different `git.exe` files. The running one is `/clangarm64/bin/git.exe`, and the signed one is `/usr/src/MINGW-packages/src/git/git.exe`. So that's not the problem.

dscho:
> Since you're RDP'ing into the machine anyway, could you maybe collect a full memory dump of a hanging process using [procdump](https://learn.microsoft.com/en-us/sysinternals/downloads/procdump)? Maybe that could shed some more light on what it's waiting for 🤔

@Alovchin91 I am curious what you suggest to do with that dump. Sure, I can collect that information, but what does that buy us that the previous investigation ([#1](https://github.com/git-for-windows/git-for-windows-automation/pull/22#issuecomment-1433600039) and [#2](https://github.com/git-for-windows/git-for-windows-automation/pull/22#issuecomment-1433111102)) didn't reveal yet?

@dennisameling @jeremyd2019 did any of you manage to reproduce this in a non-CI setting, i.e. when starting a command interactively? I tried and failed to cause such a hang using interactive commands in an Azure VM, and would love to do that to accelerate the investigation.

Alovchin91:
> @Alovchin91 I am curious what you suggest to do with that dump. Sure, I can collect that information, but what does that buy us that the previous investigation ([#1](https://github.com/git-for-windows/git-for-windows-automation/pull/22#issuecomment-1433600039) and [#2](https://github.com/git-for-windows/git-for-windows-automation/pull/22#issuecomment-1433111102)) didn't reveal yet?

That would give some starting point to those who would like to help but cannot reproduce. From your very next sentence I understand that reproducing the issue is not a trivial task 😅  I've [asked around on Mastodon](https://mastodon.social/@alovchin91/109882560933935648) if somebody could help. That, of course, if you're looking for any help 😊

jeremyd2019:
The hangs that I experienced with pacman (both signature verification and info database updates) would happen frequently while updating/installing packages interactively.  I know I did try the process dump option in task manager, but that dump was also 'missing' the context for the main thread.

From my anecdotal observations, I think this seems more likely to occur on slower machines (happened all the time on the raspberry pi, but was more rare on the QC710), but perhaps more likely with more cores? (I was playing with a 60 core VM and I think it was more likely there than with 8 core QC710).

I am set up to run a KVM virtual machine on a raspberry pi, I can try to reproduce the hangs I got with pacman and try things, if anybody has any ideas to try (I can try procdump and post that somewhere if somebody thinks it would help, for example).

dscho:
> The hangs that I experienced with pacman (both signature verification and info database updates) would happen frequently while updating/installing packages interactively.

@jeremyd2019 did you find a _minimal_ reproducer there, i.e. ideally a command that only verifies the signature but does not necessarily update any package?

Also, do you happen to know whether the signature verification happens to use a _MINGW_ program to perform that verification (e.g. `/clangarm64/bin/gpg.exe`)? In my cases, the hang always happened in the MSYS process associated with spawning a non-MSYS program...

dscho:
@Alovchin91 the thing that makes me most dubious about the `procdump` suggestion is that we _already_ attached to the process, using `gdb.exe`, and that neither `procdump` nor any other regular Win32 tools have any idea about the MSYS2 runtime or about GCC's DWARF debug symbols and are therefore even less likely than `gdb.exe` to show any useful information.

And FWIW there is already somebody from Microsoft helping this here ticket: me.

jeremyd2019:
> @jeremyd2019 did you find a _minimal_ reproducer there, i.e. ideally a command that only verifies the signature but does not necessarily update any package?

No, when I tried I ended up reproducing a *different* hang that also happened on x64, and was able to be debugged and was fixed in Cygwin, so it wasn't a complete loss, but after that was fixed my minimal reproducer no longer reproduced the hang.

> Also, do you happen to know whether the signature verification happens to use a _MINGW_ program to perform that verification (e.g. `/clangarm64/bin/gpg.exe`)? In my cases, the hang always happened in the MSYS process associated with spawning a non-MSYS program...

No, it uses msys2 gpg, via gpgme.

jeremyd2019:
> @Alovchin91 the thing that makes me most dubious about the `procdump` suggestion is that we _already_ attached to the process, using `gdb.exe`, and that neither `procdump` nor any other regular Win32 tools have any idea about the MSYS2 runtime or about GCC's DWARF debug symbols and are therefore even less likely than `gdb.exe` to show any useful information.

I tried gdb, windbg, the task manager process dump thing, even local kernel debugging, but was never able to get user mode context for that thread.

I *think* I heard that windbgx (the new version in the store) now supports DWARF symbols, but I haven't verified that myself, and I don't know if it would understand the split debug info that msys2-runtime uses.

dscho:
> when I tried I ended up reproducing a _different_ hang that also happened on x64, and was able to be debugged and was fixed in Cygwin, so it wasn't a complete loss, but after that was fixed my minimal reproducer no longer reproduced the hang.

@jeremyd2019 that's interesting. It probably means that the hangs are related, at least it could give us a better idea which code path to look at. Would you happen to know the commit hash of the fix in Cygwin?

> I _think_ I heard that windbgx (the new version in the store) now supports DWARF symbols, but I haven't verified that myself, and I don't know if it would understand the split debug info that msys2-runtime uses.

Indeed. That's why I think attaching with `gdb.exe` is our best bet to get any useful stack trace.

But I'm starting to believe that this is a dead end, we might never get a useful stack trace because the issue is more likely that something in MSYS2's tear-down code is failing to tear down the signal event handler thread, and that one is blocking the correct and expected termination of the process.

So I fear that we'll need to work with heavy instrumentation ("debug small_printf() messages littered all over the tear-down code path") and then compare the logs of a hanging to a non-hanging invocation of any minimal reproducer that we can come up with.

dscho:
> Would you happen to know the commit hash of the fix in Cygwin?

@jeremyd2019 this might actually help address the hangs I observed, as Git for Windows is still on the Cygwin v3.3.* train.

jeremyd2019:
https://github.com/msys2/msys2-runtime/commit/0ce992c1e40f843d5f264e87908a6decfd6681d1

dscho:
@jeremyd2019 thank you. Unfortunately my hope that this could help Git for Windows is dispelled by the fact that this commit made it into v3.3.4, and we're on v3.3.6 already, so we _have_ that fix...

dscho:
Update: I managed to reproduce the hang interactively, in my Azure VM, not on the first attempt, but the second. With a self-signed certificate and the `/clangarm64/bin/git.exe` of git-for-windows/git-sdk-arm64, I ran this:

```sh
git -c alias.signtool='!sh "/usr/src/build-extra/signtool.sh"' signtool $PWD/g.exe
```

where `/usr/src/build-extra/signtool.sh` comes from [here](https://github.com/git-for-windows/build-extra/blob/main/signtool.sh) and `g.exe` was the copy of a just-built `git.exe` (and now-already signed, due to the first attempt).

The self-signed certificate was created via:

```powershell
$cert = New-SelfSignedCertificate -DnsName www.yourwebsite.com -Type CodeSigning -CertStoreLocation Cert:\CurrentUser\My
$CertPassword = ConvertTo-SecureString -String "my_passowrd" -Force -AsPlainText
Export-PfxCertificate -Cert (Get-ChildItem Cert:\CurrentUser\My -CodeSigningCert)[0] -FilePath ".sig\codesign.p12" -Password $CertPassword
"my_passowrd" |  Out-File -NoNewline -Encoding ascii ".sig\codesign.pass"
```

In a second MinTTY, when I run `ps`, I see this line:

```
     2129       1    2128       8176  pty0      197108 20:16:51 /usr/bin/sh <defunct>
```

The full output of that `ps` invocation is:

```
      PID    PPID    PGID     WINPID   TTY         UID    STIME COMMAND
     2021       1    2021      10056  ?         197108 20:14:27 /usr/bin/mintty
     2129       1    2128       8176  pty0      197108 20:16:51 /usr/bin/sh <defunct>
     2133    2132    2133       4132  pty1      197108 20:17:21 /usr/bin/bash
     2132       1    2132       9120  ?         197108 20:17:20 /usr/bin/mintty
     2217    2133    2217       1320  pty1      197108 20:28:19 /usr/bin/ps
     2022    2021    2022       8444  pty0      197108 20:14:27 /usr/bin/bash
     2128    2022    2128       6172  pty0      197108 20:16:51 /clangarm64/bin/git
```

Note that the `pty0` processes refer to the hanging session, the `pty1` processes are the ones running in the second MinTTY window where I run `ps`.

The really strange thing about this is that no Win32 process has ProcessId 8176 (remember: the MSYS2/Cygwin runtime maintains its own pid that is different from the actual Win32 ProcessId), but there is no such process when I look via `wmic process where parentprocessid=8176 get ProcessId,ParentProcessId,CommandLine`.

The parent process (MSYS2/Cygwin pid 2128, corresponding to the Win32 ProcessId 6172) refers to the `git.exe` process that I called, i.e. that `sh` should be the one that ran the `signtool.sh` script. But when I look for such a process, I see a _different_ Win32 ProcessId:

```console
$ wmic process where 'CommandLine like "%signtool%"' get ProcessId,ParentProcessId,CommandLine
CommandLine                                                                                                                                                                ParentProcessId  ProcessId
C:\git-sdk-arm64\clangarm64\bin\git.exe -c "alias.signtool=!sh \"/usr/src/build-extra/signtool.sh\"" signtool C:/git-sdk-arm64/usr/src/MINGW-packages/mingw-w64-git/g.exe  5652             6172
sh -c "sh \"/usr/src/build-extra/signtool.sh\" \"$@\"" "sh \"/usr/src/build-extra/signtool.sh\"" C:/git-sdk-arm64/usr/src/MINGW-packages/mingw-w64-git/g.exe               6172             6692
C:\Windows\System32\Wbem\wmic.exe process where "CommandLine like \"%signtool%\"" get ProcessId,ParentProcessId,CommandLine                                                6244             3288
```

The Win32 ProcessId of that `sh` process is 6692, and there is _no_ line in the `ps` output corresponding to that process.

Another strange thing is that the `ps` output says that the parent process of that `git.exe` process has Win32 ProcessId 8444, but the `wmic` output suggests that the parent process has Win32 ProcessId 5652. So I looked for that process and it seems that it is running, with the parent process being the one with Win32 ProceessId 8444 and the same cmdline as that:

```console
$ wmic process where processid=5652 get ProcessId,ParentProcessId,CommandLine   CommandLine                                   ParentProcessId  ProcessId
C:\git-sdk-arm64\usr\bin\bash.exe --login -i  8444             5652
```

Alovchin91:
> And FWIW there is already somebody from Microsoft helping this here ticket: me.

Oh, I didn't know that bit, sorry 😅 I was just trying to be helpful 😬

I've been thinking about procdump because it can create a _full_ dump. AFAIK task manager creates a minidump. My understanding is that it can be analysed by any tool afterwards. But that was just a suggestion.

jeremyd2019:
Well, here's a full procdump of a hung bash.exe, if anyone wants to try to learn anything from it.

[bash.exe_230218_214907.dmp.gz](https://github.com/git-for-windows/git-for-windows-automation/files/10775257/bash.exe_230218_214907.dmp.gz)

dscho:
FWIW I could reproduce the hang on my Azure VM only once, and then no more :-(

jeremyd2019:
You might try putting the system under load while trying to reproduce it (trying to replicate my 'slower machines' observation).  This was involved when I was trying to come up with a reproducer (that wound up finding that other bug), I was basically stressing fork/exec hoping it would eventually hang.

PR #25, opened 2023-02-07
build-and-deploy: several optimizations, mostly to facilitate arm64 workflows

dennisameling:
Successful run that includes these changes: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4114193682

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -276,12 +276,6 @@ jobs:
                     esac
           )
 
-          # No need to rebuild i686/x86_64 flavors when building the initial
-          # arm64 versions of the wintoast/GCM packages
-          test ebe6c944c573ee4d8900f0e015365468606c9e1e != "$REF" ||
-          test "mingw32 mingw64 clangarm64" != "$MINGW_ARCHS_TO_BUILD" ||
-          MINGW_ARCHS_TO_BUILD=clangarm64
-
           cd "/usr/src/$REPO/$PACKAGE_TO_BUILD" &&
           MAKEFLAGS=-j6 PKGEXT='.pkg.tar.xz' MINGW_ARCH=$MINGW_ARCHS_TO_BUILD $MAKEPKG -s --noconfirm &&
           cp *.pkg.tar* "$dir/" &&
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -250,9 +250,20 @@ jobs:
...
+            *)
+              GIT_EXE="/mingw64/bin/git.exe"
+              BUILD_SRC="YesPlease"
+              ;;
+          esac &&
+
           {
             test -f /usr/bin/git ||
-            printf '#!/bin/sh\n\nexec /mingw64/bin/git.exe "$@"\n' >/usr/bin/git
+            printf '#!/bin/sh\n\nexec '$GIT_EXE' "$@"\n' >/usr/bin/git
```
rimrul, line 264: I think we could theoretically use `$MINGW_PREFIX/bin/git.exe` instead of setting `$GIT_EXE` in that `case`.

.github/workflows/build-and-deploy.yml
```diff
@@ -250,9 +250,20 @@ jobs:
...
+          case "$ARCHITECTURE" in
+            aarch64)
+              GIT_EXE="/clangarm64/bin/git.exe"
+              BUILD_SRC=
+              ;;
+            *)
+              GIT_EXE="/mingw64/bin/git.exe"
+              BUILD_SRC="YesPlease"
+              ;;
+          esac &&
```
dscho, line 260: Shouldn't we have an `x86_64` arm and an `i686` arm in this `case` statement, and a `*)` arm that errors out?

Also, here would be a fine spot to define `MINGW_PREFIX`, probably instead of `GIT_EXE`.

Lastly, given how much time it takes to build the source package, maybe we should consider yet another "ARCHITECTURE" called `sources`?

.github/workflows/build-and-deploy.yml
```diff
@@ -276,12 +276,6 @@ jobs:
                     esac
           )
 
-          # No need to rebuild i686/x86_64 flavors when building the initial
-          # arm64 versions of the wintoast/GCM packages
-          test ebe6c944c573ee4d8900f0e015365468606c9e1e != "$REF" ||
-          test "mingw32 mingw64 clangarm64" != "$MINGW_ARCHS_TO_BUILD" ||
-          MINGW_ARCHS_TO_BUILD=clangarm64
-
```
dscho, line 273: ❤️

.github/workflows/build-and-deploy.yml
dscho: > The `build-and-deploy` pipeline is currently designed in a way that it uses a single `windows-latest` runner (`x86_64`) to build both `mingw32` and `mingw64` packages in one go. That's why I set up the logic this way.

Oh whoops. I totally was thinking about the `git-artifacts` workflow when I said this last night.

In light of that, even the `MINGW_PREFIX` suggestion does not make sense to me any longer, because there would have to be _two_ such prefixes in the non-ARM64 case.

So I retract my suggestions 😁

dscho (approved):

PR #30, opened 2023-02-10
self-hosted-runners: use 4-core VMs

dennisameling:
In a recent Actions run, we ran into the CPU quotum that was set in Azure (max. 20 cores across all active VMs). By lowering the vCPU count on the VMs from 8 to 4, we can increase the number of simultaneous VMs from 2 to 5.

Next to that, the current performance is mostly limited by x64-emulation of unixy-tools like Bash anyway. The workflows really aren't using all the CPU cores, only during things like Clang-compilation.

Ref: https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4144779320/jobs/7168260908


dscho (approved):
Great!

We _may_ want to consider teaching the GitForWindowsHelper GitHub App to pay attention to the total number of currently-running Windows/ARM64 jobs and hold off creating new VMs accordingly (probably using successful `delete-self-hosted-runner` jobs as trigger to see whether a held-off creation needs to be started).

But that sounds a bit complex to set up and make robust, so let's only do this if we really need to (read: if we run into this so often that manual VM creations become too tedious).

PR #31, opened 2023-02-11
open-pr: don't concatenate  after last command of a step

rimrul:
We've split a step into two, but left the `&&` at the end of the first. That leads to failures like this one:

https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4150668809/jobs/7180481501


dscho (approved):
Obviously correct!

PR #34, opened 2023-02-15
git-artifacts: allow building from an existing tag

dscho:
So far, we have an implementation of the common flow where the `tag-git` workflow is started to create "bundle-artifacts" including a Git bundle with the newly-tagged Git version and another Git bundle with the updated release notes.

However, for the just-published Git for Windows v2.39.2, we would like to build the ARM64 flavor of the artifacts, i.e. for an _existing_ tag.

Let's introduce two new inputs (referring to an existing tag and an existing build-extra revision) that can be used instead of the input referring to a `tag-git` workflow run, and handle that scenario, too.

### DIFF
--- a/.github/workflows/git-artifacts.yml
+++ b/.github/workflows/git-artifacts.yml
@@ -26,6 +26,12 @@ on:
       bundle_artifacts_workflow_run_id:
         description: 'Workflow run ID from bundle-artifacts pipeline'
         required: false
+      existing_git_tag:
+        description: 'Existing tag to build from. Requires bundle_artifacts_workflow_run_id to be empty'
+        required: false
+      build_extra_rev_for_existing_git_tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/git-artifacts.yml
```diff
@@ -7,14 +7,6 @@ on:
...
         description: 'Optionally restrict what artifacts to build (portable, installer, etc.). Separate artifacts with spaces'
         required: false
-      ref:
-        description: 'Optionally override which branch to build'
-        required: false
-        default: main
-      repository:
-        description: 'Optionally override from where to fetch the specified ref'
-        required: false
-        default: git-for-windows/git
```
dennisameling, line 17: Thanks for removing those! I noticed them earlier this week (copied over from the original pipeline) and was planning to remove them indeed :)

dennisameling (approved):
Thank you! 🚀

Issue #36, opened 2023-02-21
build-and-deploy: do not skip the `refresh` step


dscho:
In https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4231051093/jobs/7349093571#step:22:49, we see what happens if, say, the `pacman` step fails and the `refresh` step is skipped: we try to update the check run but fail to refresh the token that is necessary to do that.

In [another run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4231048883/jobs/7349088841#step:22:37) the symptom looks different: it is indicated that we run into API rate limiting. That's because the used token is empty and therefore we run unauthenticated.

Fix that by forcing the `refresh` step just like the `update check-run` step.

dscho:
https://github.com/git-for-windows/git-for-windows-automation/pull/38 made this obsolete, as it removed the `refresh` step.

PR #38, opened 2023-02-28
Adjust the `tag-git` workflow and create a `release-git` workflow to build Git for Windows fully using GitHub Actions

dscho:
- [x] adjust the `tag-git` workflow so that it can tag full Git for Windows versions, not just snapshots
- [x] adjust the `git-artifacts` workflow to mirror Check Runs to `git-for-windows/git` (needs to encode the architecture as part of the Check Run name to avoid them overwriting each other)
- [x] reimplement [the Azure Release Pipeline](https://dev.azure.com/git-for-Windows/git/_release?definitionId=1&_a=releases) as new GitHub workflow, called `release-git`

This addresses https://github.com/git-for-windows/git-for-windows-automation/issues/37

PR #39, opened 2023-03-04
build-and deploy: refresh token, even on failure

rimrul:
the "mark check run as completed" step depends the token output of this step.

This fixes https://github.com/git-for-windows/git-for-windows-automation/issues/36


rimrul:
#38 makes this obsolete, as it removes the refresh step

PR #42, opened 2023-03-17
Fix pushing `MINGW-packages` updates in the `release-git` workflow

dscho:
While the `release-git` workflow run looked as if it performed all the repository updates as expected, it actually only fetched from the `MINGW-packages.bundle` but did not merge the change. As a consequence, the `git push` had no updates to offer.

I fixed that by manually downloading the bundle in [the `git-artifacts (x86_64)` run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4413112845#artifacts)'s `pkg-x86_64` artifact and pushing https://github.com/git-for-windows/MINGW-packages/commit/13a189253e9fef04e18fdbd51254930905c3b9b1.

However, it is a better idea to fix this in the `release-git` workflow than to do that manual fix-up for every release.

### DIFF
--- a/repository-updates.js
+++ b/repository-updates.js
@@ -20,6 +20,33 @@ const getWorkflowRunArtifact = async (context, token, owner, repo, workflowRunId
   await downloadAndUnZip(token, urls[name], name)
 }
 
+const mergeBundle = (gitDir, worktree, bundlePath, refName) => {
+  callGit(['--git-dir', gitDir, 'fetch', bundlePath, refName])
+  // If there is nothing to merge, return early
+  if (callGit(['--git-dir', gitDir, 'rev-list', '--count', `${refName}..FETCH_HEAD`]) === '0') return
+
+  if (worktree) {
+    callGit(['branch', '-M', refName], worktree)
+    callGit(['merge', '--no-edit', 'FETCH_HEAD'], worktree)
+  } else {
+    // If it fast-forwards, we do not need a worktree
+    if (callGit(['--git-dir', gitDir, 'rev-list', '--count', `FETCH_HEAD..${refName}`]) === '0') {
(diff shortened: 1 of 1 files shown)


dscho:
@rimrul could I solicit your review? My plan is to exercise this code in the next `-rc0` already ([scheduled for May 12th](https://gh.io/gitCal)) and I would therefore like to merge this PR before then.

repository-updates.js
```diff
@@ -20,6 +20,33 @@ const getWorkflowRunArtifact = async (context, token, owner, repo, workflowRunId
...
+  } else {
+    // If it fast-forwards, we do not need a worktree
+    if (callGit(['--git-dir', gitDir, 'rev-list', '--count', `FETCH_HEAD..${refName}`]) === '0') {
+      callGit(['--git-dir', gitDir, 'update-ref', `refs/heads/${refName}`, 'FETCH_HEAD'])
+    } else {
+      // Fine. We need a worktree. But we don't have one. So let's create one.
+      worktree = `${gitDir}/tmp-worktree`
+      callGit(['--git-dir', gitDir, 'worktree', 'add', '--no-checkout', worktree])
+      // No need for a full checkout, it's not as if we want to resolve any merge conflicts...
+      callGit(['sparse-checkout', 'init'], worktree)
```
vdye, line 40:
```suggestion
      callGit(['sparse-checkout', 'set'], worktree)
```

`sparse-checkout init` has been deprecated, but you can use `sparse-checkout set` to get the same result.

dscho: [Fixed](https://github.com/git-for-windows/git-for-windows-automation/compare/2d306abda1ee728c8294b889610fe762f9d0fbe6..65eb988f268e9b376d05624bdb7a23257d887c9e)!

PR #43, opened 2023-03-17
release-git: update the `release` branch if necessary

dscho:
We already have a check to verify that the `release` branch is up to date with the `main` branch before even starting the `release-git` workflow run for real.

But we can do better than telling the user when the branch is not up to date: We can easily try to update it in the workflow run ;-)

### DIFF
--- a/.github/workflows/release-git.yml
+++ b/.github/workflows/release-git.yml
@@ -40,7 +40,20 @@ jobs:
             })
 
             if (sha !== '${{ github.sha }}') {
-              throw new Error(`The 'release' branch is not up to date!
+              // Try to update the `release` branch by fast-forwarding to `main`
+              try {
+                const { pushGitBranch } = require('./repository-updates.js')
+                await pushGitBranch(
+                  console,
+                  core.setSecret,
+                  ${{ secrets.GH_APP_ID }}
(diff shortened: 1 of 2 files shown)


dscho:
For the record, I plan on testing this while releasing the next `-rc0`.

vdye (commented):
The `release` vs. `main` setup is quite confusing, but I _think_ I understand what it's doing. Assuming that understanding is correct, one general question I have is - have the "run job, stage fails, update `release`, then re-run failed jobs" scenario been tested? In the past, I've encountered unexpected issues when changing a ref under-the-hood and re-running a job using that ref. 

Otherwise

dscho:
> The `release` vs. `main` setup is quite confusing, but I _think_ I understand what it's doing.

Yes, I understand. It is the "least worst" solution I found to a problem that has no proper solution.

> Assuming that understanding is correct, one general question I have is - have the "run job, stage fails, update `release`, then re-run failed jobs" scenario been tested?

Yes, it has! [Here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4332770958):

image


and [here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/4413389593):

image

vdye (approved):

PR #47, opened 2023-05-17
Fixes for check run updates

dscho:
The [latest `git-artifacts` run failed](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/5001089177) because the installation token was created more than 10 minutes before the `portable` and the `nuget` jobs were actually started (they were queued for a long time because of some concurrent workflow runs using up the contingent of hosted runners available to the Git for Windows project).

This uncovered the fact that we fail to re-generate new installation tokens because the Action lacks the proper information: it needs a repository to work with, but that information was not passed as parameter.

This is a wide-spread issue: the `check-run-action` actually requires the owner/repo information in case a new installation token needs to be generated.

This PR fixes all of those instances, and while in the space also ties up a few other loose ends.

### DIFF
--- a/.github/actions/check-run-action/action.yml
+++ b/.github/actions/check-run-action/action.yml
@@ -9,17 +9,13 @@ inputs:
     required: true
   owner:
     description: 'The owner of the target repository'
-    required: true
     default: 'git-for-windows'
   repo:
     description: 'The name of the target repository'
-    required: true
   rev:
     description: 'The target commit'
-    required: true
   check-run-name:
     description: 'The name of the Check Run'
-    required: true
   title:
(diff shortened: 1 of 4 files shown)


dscho:
Hmm. After having another look, I think my fix is wrong. The idea was, after all, to store the relevant information in the `state` file, including the owner and the repository.

dscho:
There you go. That's the ticket. The owner/repo information _was_ stored in the state file, but I simply forgot to use it 😄. _Now_ this PR is ready to be merged.

dscho:
I'll just merge this before building another set of Git artifacts over in https://github.com/git-for-windows/git/pull/4426.

PR #63, opened 2024-01-03
Return non-zero status from update scripts on failure

### LINKED ISSUE #53: update-scripts should return non-zero value on failure
See https://github.com/git-for-windows/git-for-windows-automation/actions/runs/5672350769/job/15371376778#step:13:42 for an example of an `open-pr` run that continued "successfully" after failing to update the checksums of the at the time non-existing `git-lfs` release assets.

rimrul:
This fixes #53


dscho (approved):
Thank you so much!

PR #73, opened 2024-02-20
github-release: in addition to the tag, also push `main`

dscho:
The idea is always to update the `main` branch after the Git for Windows release is done, to close the PR.

So far, this is done manually. But it's better to automate this. Because then I cannot make mistakes like merging via the `Merge` button (as I had done for v2.44.0-rc1, which complicated subsequent rebases) instead of pushing to `main`.

The downside would be that the PR would be closed before the `release` workflow run had a chance to finish, and therefore any failures may be harder to spot.

### DIFF
--- a/.github/actions/github-release/action.yml
+++ b/.github/actions/github-release/action.yml
@@ -77,7 +77,7 @@ runs:
             checkSHA256Sums,
             createRelease,
             uploadGitArtifacts,
-            pushGitTag,
+            pushGitTagAndMainBranch,
             updateRelease
           } = require('./github-release')
 
@@ -126,7 +126,7 @@ runs:
           )
           await state.refreshToken()
 
-          await pushGitTag(
+          await pushGitTagAndMainBranch(
             console,
             core.setSecret,
             state.accessToken,
--- a/github-release.js
+++ b/github-release.js
@@ -244,7 +244,7 @@ const uploadGitArtifacts = async (context, token, owner, repo, releaseId, gitArt
   context.log('Done uploading Git artifacts')
 }
 
-const pushGitTag = (context, setSecret, token, owner, repo, tagName, bundlePath) => {
+const pushGitTagAndMainBranch = (context, setSecret, token, owner, repo, tagName, bundlePath) => {
   context.log(`Pushing Git tag ${tagName}`)
   const { callGit } = require('./repository-updates')
   callGit(['clone',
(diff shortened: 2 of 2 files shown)


dscho:
> The downside would be that the PR would be closed before the `release` workflow run had a chance to finish, and therefore any failures may be harder to spot.

I guess a much better idea is to do this in the [GitForWindowsHelper GitHub App](https://github.com/git-for-windows/gfw-helper-github-app), listening for a successful `release` Check Run on the respective `git-for-windows/git` PR, [much like it listens for successful `tag-git` runs](https://github.com/git-for-windows/gfw-helper-github-app/blob/c5ebb1c032f4c4fd1f97b603b9dc52e3ea9b3dc7/GitForWindowsHelper/cascading-runs.js#L112-L128).

The only slightly tricky bit is that there are actually _several_ Check Runs, not just a single one:
- `github-release`
- `announcement-mail` (depends on `github-release`)
- `gitforwindows.org` (depends on `github-release`, is intentionally skipped for -rc versions)
- `repository-updates` (depends on `github-release`, is intentionally skipped for -rc versions)
- `pacman-updates`
- `nuget-packages`

All of these would need to be checked via the REST API which requires a bit of pagination (e.g. https://api.github.com/repos/git-for-windows/git/commits/0db3af6cead6855bb0d72b7149f5c2d9adb5f835/check-runs?per_page=100&page=2).

An alternative would be to access the `git-for-windows-automation` workflow run (via the link in the output summary) and look at the jobs (see e.g. https://api.github.com/repos/git-for-windows/git-for-windows-automation/actions/runs/7978659901/jobs). The problem is that those might not have finished yet.

A better alternative, now that I think of it, would be to listen for the completed _workflow_ run in the `git-for-windows-automation` repository. There, we simply have the conclusion that it was a success, because all the jobs run (and are not skipped, it's just some mirrored Check Runs that are marked as skipped for -rc versions). The trick would be to get from that workflow run to the corresponding `git-for-windows/git` PR because there is no information in the Get a Check Run response to get back to the `git-for-windows/git` PR or tag or commit (example: https://api.github.com/repos/git-for-windows/git-for-windows-automation/actions/runs/7978659901). Sadly, searching for a PR comment containing the workflow run ID won't work because the REST API search endpoint searches the rendered text, not the Markdown. But we can [search for an open PR with the comment](https://api.github.com/search/issues?q=org:git-for-windows%20is:pr%20is:open%20%22The%20release-git%20workflow%20run%20was%20started%22%20in:comments) and then follow the `comments_url` (e.g. https://api.github.com/repos/git-for-windows/git/issues/4828/comments) that _does_ have that link _and_ has `author_association == 'MEMBER'` (so that we don't fall for jokers who open bogus PRs and add a comment of the exact correct format to throw the automation).

The GitForWindowsHelper GitHub App would need to use [the `repos/.../git/refs/main` endpoint](https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#update-a-reference) endpoint (with `force=false`, to avoid accidental force-pushing).

PR #81, opened 2024-06-05
workflows: add timeouts, address linter issues

dscho:
Last night, a `mingw-w64-openssl (aarch64)` job [timed out after 6h](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/9374447637/job/25810488159#step:12:28), which cost quite a bit of Azure resources.

The reason was most likely that finicky MSYS2 runtime which hangs on Windows/ARM64 quite often.

Let's add timeouts to workflows' steps to avoid incurring these Azure costs.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -162,6 +162,7 @@ jobs:
           bash -lc "set -x && rebase -v -i /usr/lib/perl5/core_perl/auto/Cwd/Cwd.dll"
 
       - name: Get GPG key(s)
+        timeout-minutes: 5
         shell: bash
         env:
           CARCH: x86_64 # dummy, to allow sourcing cv2pdb's PKGBUILD as-is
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -204,6 +205,7 @@ jobs:
           echo "SIGNTOOL=git signtool" >>$GITHUB_ENV
 
       - name: Build ${{env.PACKAGE_TO_BUILD}}
+        timeout-minutes: 150
```
rimrul, line 208: This one needs a special case for LLVM.
 something like `timeout-minutes ${{ env.PACKAGE_TO_BUILD== 'mingw-w64-llvm' && '
300 || 150 }}`

dscho: Done!

rimrul (approved):

PR #83, opened 2024-06-10
Fix the `updpkgsums` GitHub workflow

dscho:
I never had verified that the `/updpkgsums` slash command works, up until last week, when I tried to run it in git-for-windows/build-extra#558. Turns out that it needed a couple of fixes. https://github.com/git-for-windows/gfw-helper-github-app/pull/83 was needed on the GitHub App side, and this here PR is needed to actually let the operation succeed.

### DIFF
--- a/.github/actions/init-g4w-sdk-for-pacman/action.yml
+++ b/.github/actions/init-g4w-sdk-for-pacman/action.yml
@@ -1,5 +1,10 @@
 name: 'Initialize Git for Windows SDK subset to run `pacman`'
 description: 'This composite GitHub Action initializes a subset of the Git for Windows SDK intended to run `pacman` and friends'
+inputs:
+  include-makepkg:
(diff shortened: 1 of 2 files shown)


dscho:
[Here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/9455615871/job/26045785895) is a successful run that verifies that no messages like these are shown:

```
==> Making package: mingw-w64-git-extra 1.1.636.2db97b993-1 (Fri Jun  7 12:41:16 2024)
==> Checking runtime dependencies...
error: GPGME error: Invalid crypto engine
error: clangarm64: missing required signature
[...]
```

These would be shown unless a working `gpg.exe` is part of the SDK subset, triggered via `makepkg` running `pacman -T diffutils`, among other, similar `pacman` invocations.

Took a while to figure out how best to include the missing bits, but finally I managed that trick.

PR #84, opened 2024-06-24
self-hosted runners: upgrade to 24h2

dennisameling:
The 24h2 version of Windows 11 introduces a new x64 emulation engine: Prism. It promises improved performance, and maybe it'll help resolve some of the mysterious arm64 hangs we've been experiencing...

Ref: https://learn.microsoft.com/en-us/windows/arm/apps-on-arm-x86-emulation#prism

### DIFF
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -39,7 +39,7 @@ env:
   # For more information, see https://learn.microsoft.com/en-us/azure/virtual-machines/dplsv5-dpldsv5-series (which
   # unfortunately does not have more information about price by region)
   AZURE_VM_REGION: westus2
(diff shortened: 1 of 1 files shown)


dennisameling:
For testing this, I'd like to kick off 10 pipeline runs of `git-artifacts` for `aarch64` using the existing `23h2` version, and another 10 using the `24h2` version. We've been experiencing some hangs in the code signing process there, and I'm curious if the new emulation engine (Prism) will resolve those issues at all. WDYT?

dscho:
> I'm curious if the new emulation engine (Prism) will resolve those issues at all. WDYT?

Given [this comment](https://github.com/msys2/msys2-autobuild/issues/62#issuecomment-2188011588), I would not expect Prism to work around the issue 😞

dscho (approved):
Nice, thank you!

PR #85, opened 2024-07-02
release: create a discussion for every Git for Windows release

dscho:
It is a nice thing to have for users, and therefore I just created one for the latest Git for Windows version at
https://github.com/git-for-windows/git/discussions/5035.

Let's automate this.

For details about the used REST API endpoint, see
https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release--parameters

### DIFF
--- a/github-release.js
+++ b/github-release.js
@@ -9,6 +9,7 @@ const createRelease = async (context, token, owner, repo, tagName, rev, name, bo
       target_commitish: rev,
       name,
       body,
+      discussion_category_name: 'Announcements',
       draft: draft === undefined ? true : draft,
       prerelease: prerelease === undefined ? true : prerelease
     }


rimrul (approved):

rimrul:
It looks like this didn't work for 2.46.0.

dscho:
> It looks like this didn't work for 2.46.0.

Hmm. Strange. Maybe it's because this is in `createRelease()`, which is called to create the draft release only, so that the assets can be uploaded.

dscho:
I created [the discussion](https://github.com/git-for-windows/git/discussions/5097) manually, using:

```sh
gh api -X PATCH repos/git-for-windows/git/releases/167714326 -F discussion_category_name='Announcements'
```



> Maybe it's because this is in `createRelease()`, which is called to create the draft release only, so that the assets can be uploaded.

I think this is the most plausible explanation. The information is simply not retained when creating the release in draft mode, and [when the release is published](https://github.com/git-for-windows/git-for-windows-automation/blob/98aad93fdc9b9136deb97e407379f2a7c23860b7/.github/actions/github-release/action.yml#L142-L152), that attribute is forgotten and no discussion is created.

So I think we'll need this:

```diff
diff --git a/.github/actions/github-release/action.yml b/.github/actions/github-release/action.yml
index 9d925f0..3e0bbee 100644
--- a/.github/actions/github-release/action.yml
+++ b/.github/actions/github-release/action.yml
@@ -147,7 +147,8 @@ runs:
             release.id, {
               draft: false,
               prerelease: isPrerelease,
-              make_latest : !isPrerelease
+              make_latest : !isPrerelease,
+              discussion_category_name: 'Announcements',
             }
           )
     - name: update check-run
```

Issue #40, opened 2023-03-07
Add a workflow to run `pacman-helper break-lease


dscho:
This is a companion to https://github.com/git-for-windows/build-extra/pull/488 with the intention of introducing a `/break pacman-upload-lease` command in https://github.com/git-for-windows/gfw-helper-github-app.

PR #86, opened 2024-07-04
Add a workflow to break a pacman upload lease

rimrul:
This fixes https://github.com/git-for-windows/git-for-windows-automation/issues/40

### DIFF
--- a/.github/workflows/break-pacman-upload-lease.yml
+++ b/.github/workflows/break-pacman-upload-lease.yml
@@ -1,0 +1,35 @@
+name: break-pacman-upload-lease
+run-name: Break pacman upload lease
+
+on:
+  workflow_dispatch:
+
+jobs:
+  break-lease:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone build-extra
+        shell: bash
+        run: git clone --depth 1 --single-branch -b main https://github.com/git-for-windows/build-extra /usr/src/build-extra
+
+      - name: Prepare home directory
+        if: env.AZURE_BLOBS_TOKEN != ''
+        env:
+          AZURE_BLOBS_TOKEN: ${{secrets.AZURE_BLOBS_TOKEN}}
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


rimrul:
This could probably almost run on `ubuntu-latest` runner, but we use `cygpath` in these scripts.

.github/workflows/break-pacman-upload-lease.yml
```diff
@@ -0,0 +1,40 @@
...
+  workflow_dispatch:
+
+jobs:
+  break-lease:
+    runs-on: windows-latest
+    steps:
+      - uses: git-for-windows/setup-git-for-windows-sdk@v1
+        with:
+          flavor: minimal
+          architecture: x86_64
```
dscho, line 14: Do we even need the minimal SDK? As far as I can tell, [`pacman-helper.sh break_lock`](https://github.com/git-for-windows/build-extra/blob/HEAD/pacman-helper.sh) simply hands off to [`wingit-snapshost-helper.sh`](https://github.com/git-for-windows/build-extra/blob/HEAD/wingit-snapshot-helper.sh), which calls `curl`. GitHub workflows' `bash` steps use Git for Windows' Bash (!!!), which is distributed with a working `curl`, so it _should_ work out of the box.

dscho (approved):
Nice, thank you!

I just have one question regarding the need to set up a minimal Git for Windows SDK. It's a minor concern, though, I'd be fine with merging as-is.

.github/workflows/break-pacman-upload-lease.yml
rimrul: We also use `date`, `xxd`, `tr`, `sed`, `base64` and `cygpath`, but that should work.

dscho:
@rimrul I opened https://github.com/git-for-windows/build-extra/pull/566 (ran it in Linux to verify that it works; It said "409 There is currently no lease on the blob." of course 😁); How do you feel about merging that first and then switching this here GitHub workflow to `ubuntu-latest`?

dscho (approved):
Yeah!

PR #89, opened 2024-08-13
github-release: do create a discussion post for each new Git for Windows version

dscho:
This was the intention of #85, but that PR only changed the call that creates the draft release.

Apparently the intention to create a discussion post is _not_ stored when creating a draft release, though, sadly. Therefore we need to state that intention again when we do publish the release.

This problem was reported [here](https://github.com/git-for-windows/git-for-windows-automation/pull/85#issuecomment-2259878251).


rimrul (approved):
If we identified the problem correctly, this should fix it.

PR #60, opened 2023-11-22
Self hosted runner: update image and make configurable through env vars

dennisameling:
Updates the self hosted runner image and also makes it configurable through env vars.

Note: the rest of this PR description was for an older version of this PR and therefore collapsed.

<details>
  <summary>See prior PR description</summary>

As discussed [here](https://github.com/git-for-windows/git-for-windows-automation/issues/59#issuecomment-1821185442), let's add some cleanup logic for Azure VMs that we no longer need.

> Another idea: we should be able to teach GitForWindowsHelper to detect failed create-self-hosted-runner runs and automatically kick off the delete run, no?

In this PR, I added logic that goes through the resource group and deletes leftover VMs. That way, it's not strictly linked to failed runs. I think this is a more robust way to ensure leftover VMs are deleted, but very much open to ideas and feedback.

> And _maybe_ also try re-running the failed run, up to five times?

By "failed run", do you mean the run that creates the self hosted runner, or the run that does the actual build of GfW ARM64? Assuming you mean the former, we'll probably have to add some logic [here](https://github.com/git-for-windows/gfw-helper-github-app/blob/main/GitForWindowsHelper/index.js) to detect when a run that creates a self hosted runner has failed, then retry, right? Not sure if that's worth the effort tbh, as these jobs have been rock-solid until I tried again yesterday after some time of not doing it. It probably deserves some more investigation to ensure that VMs are created reliably again.
</details>

PR #93, opened 2024-09-14
Clean up self hosted runaways

dscho:
Resurrect the part dropped from https://github.com/git-for-windows/git-for-windows-automation/pull/60, where we considered to add a scheduled workflow to clean up self-hosted runners that were not cleaned up as intended.

This PR adds a couple of commits on top, most notably querying GitHub whether the runner in question is still busy with something.

[EDITED] I verified in a manual test that the default scope of `$GITHUB_TOKEN` is insufficient to run that query. Therefore, we now authenticate the GitHub CLI as a GitHub App to be able to use that query.

[EDITED] Also: I changed the workflow to use [notices](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-notice-message) and [warnings](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message) to make it easier to spot what the individual workflow runs were doing.

@dennisameling would you kindly have a look?

### DIFF
--- a/.github/workflows/cleanup-self-hosted-runners.yml
+++ b/.github/workflows/cleanup-self-hosted-runners.yml
@@ -1,0 +1,56 @@
+name: Cleanup Azure self hosted runners
+run-name: Cleanup Azure self hosted runners
+
+on:
+  schedule:
+    # Run every 6 hours
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+
+# The following secrets are required for this workflow to run:
+# AZURE_CREDENTIALS - Credentials for the Azure CLI. It's recommended to set up a resource
+#                     group specifically for self-hosted Actions Runners.
+#   az ad sp create-for-rbac --name "{YOUR_DESCRIPTIVE_NAME_HERE}" --role contributor \
+#     --scopes /subscriptions/{SUBSCRIPTION_ID_HERE}/resourceGroups/{RESOURCE_GROUP_HERE} \
+#     --sdk-auth
+# AZURE_RESOURCE_GROUP - Resource group to create the runner(s) in
+jobs:
+  delete-runner:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Azure Login
+      uses: azure/login@v1
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+    - name: Discover VMs to delete
+      uses: azure/CLI@v1
+      with:
+        azcliversion: 2.54.0
+        inlineScript: |
+          active_vms=$(az vm list -g ${{ secrets.AZURE_RESOURCE_GROUP }} | jq -c '.[] | {name,timeCreated}')
+          current_time=$(date +%s)
+          three_hours_ago=$(($current_time - 3 * 3600))
+
+          if [ -z "$active_vms" ]; then
+              echo "No active VMs found, nothing to do."
+              exit 0
+          else
+              echo "Found these active VMs:"
+              echo $active_vms
+          fi
+
+          for active_vm in ${active_vms[@]}; do
+              vm_name=$(echo $active_vm | jq '.name')
+              vm_creation_iso_string=$(echo $active_vm | jq -r '.timeCreated')
+              vm_creation_time=$(date -d $vm_creation_iso_string +%s)
+              
+              if [ "$three_hours_ago" -ge "$vm_creation_time" ]; then
+                  echo "The VM ${vm_name} was created more than 3 hours ago and wasn't deleted. Let's do that now."
+                  az vm delete -n "$vm_name" -g ${{ secrets.AZURE_RESOURCE_GROUP }} --yes
+                  az network nsg delete -n "$vm_name"-nsg -g ${{ secrets.AZURE_RESOURCE_GROUP }}
+                  az network vnet delete -n "$vm_name"-vnet -g ${{ secrets.AZURE_RESOURCE_GROUP }}
+                  az network public-ip delete -n "$vm_name"-ip -g ${{ secrets.AZURE_RESOURCE_GROUP }}
+              else
+                  echo "The VM ${vm_name} was created less then 3 hours ago and shouldn't be deleted yet. Skipping."
+              fi
+          done
--- a/.github/workflows/cleanup-self-hosted-runners.yml
+++ b/.github/workflows/cleanup-self-hosted-runners.yml
@@ -19,11 +19,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Azure Login
-      uses: azure/login@v1
+      uses: azure/login@v2
       with:
         creds: ${{ secrets.AZURE_CREDENTIALS }}
     - name: Discover VMs to delete
-      uses: azure/CLI@v1
+      uses: azure/CLI@v2
       with:
         azcliversion: 2.54.0
         inlineScript: |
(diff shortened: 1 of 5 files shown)


.github/workflows/cleanup-self-hosted-runners.yml
```diff
@@ -0,0 +1,59 @@
...
+              echo $active_vms
+          fi
+
+          for active_vm in ${active_vms[@]}; do
+              vm_name=$(echo $active_vm | jq '.name')
+              vm_creation_iso_string=$(echo $active_vm | jq -r '.timeCreated')
+              vm_creation_time=$(date -d $vm_creation_iso_string +%s)
+
+              if [ "$one_hour_ago" -lt "$vm_creation_time" ]; then
+                  echo "The VM ${vm_name} was created less then 3 hours ago and shouldn't be deleted yet. Skipping."
```
dennisameling, line 48: You'll probably want to change this too ;-) 

```suggestion
                  echo "The VM ${vm_name} was created less then 1 hour ago and shouldn't be deleted yet. Skipping."
```

.github/workflows/cleanup-self-hosted-runners.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Azure Login
+      uses: azure/login@v1
+      with:
+        creds: ${{ secrets.AZURE_CREDENTIALS }}
+    - name: Discover VMs to delete
+      uses: azure/CLI@v1
+      with:
+        azcliversion: 2.54.0
```
dennisameling, line 28: We'll probably want to use the [latest](https://learn.microsoft.com/en-us/cli/azure/release-notes-azure-cli) 2.64.0, now that it's a few months later :)

dennisameling (commented):
Nice changes to the script!

.github/workflows/cleanup-self-hosted-runners.yml
dscho: Okay!

dscho: Should we do the same with the `create-azure-self-hosted-runners` and `delete-self-hosted-runner` GitHub workflows, too? They're at v2.43.0...

dscho:
> I did not test this and suspect that the default scope of `$GITHUB_TOKEN` is insufficient to run that query.

I did test this and confirm that the scope is too narrow. I've added a new script to allow authenticating as a GitHub App with `gh`, and use that instead now.

dscho:
@dennisameling I think this is now ready to be tested "in production" (meaning: trigger `create-azure-self-hosted-runners` and then temporarily switch this here workflow to `on: push` and lower the threshold of 1h, or maybe even change that threshold permanently, 40 minutes should be _plenty_ enough, no? but I'll have to run a couple of errands and may not be able to come back to this today or even tomorrow).

.github/workflows/cleanup-self-hosted-runners.yml
dennisameling: Good catch! Updated those here: https://github.com/git-for-windows/git-for-windows-automation/pull/94

Another thing I just noticed that this here PR uses `azure/CLI@v1`, while we're using `azure/CLI@v2` in the other pipelines. Another thing to update :)

dennisameling:
> I'll have to run a couple of errands and may not be able to come back to this today or even tomorrow).

I'll probably have some time to test this tomorrow. In any case, let's post a quick comment in this thread when either of us starts testing, so we don't interfere with each other.

dennisameling:
I'm starting the test now, by [creating a self-hosted runner](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10870020281). I'll let it sit for a bit and then trigger the workflow in this PR through the `workflow_dispatch` flow (which should be possible using GitHub's REST API). Otherwise I'll update the workflow to `on: push` for testing, as you suggested.

dennisameling:
The workflow [failed](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10870643871/job/30163510111) because it can't access this repo's secrets (because your PR was created from your fork instead of this repo directly).

Let me push this branch to this repo as well so we can test more easily.

dennisameling:
Looks like there was a [typo](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10870833750/job/30163936365#step:5:59) in your new `elif` statement:

```
/home/runner/work/_temp/AZ_CLI_GITHUB_ACTION_1726402015891.sh: line 29: unexpected EOF while looking for matching `''
```

Just pushed https://github.com/git-for-windows/git-for-windows-automation/commit/577630b88937dc6aa8665d3b2dfcec589ead991a to fix, as well as some follow-up commits to fix date handling.

Looks like it's [working](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10871761622/job/30165986684)! 🎉 

We'll need to clean up this PR and also add some of the [commits](https://github.com/git-for-windows/git-for-windows-automation/commits/clean-up-self-hosted-runaways/) I added to the other branch (sorry for the mess there).

I'll be afk for a couple hours now, but if you happen to have some time, please feel free to start the wrap-up of this PR. Otherwise I can probably wrap things up here tomorrow.

dscho:
Excellent, @dennisameling! I cleaned up the commits and force-pushed, and will now go ahead and merge!

dscho:
While sleeping over this, I realized that there is a chance that this workflow runs _while_ a `delete-self-hosted-runner` workflow run is still active, and in that case one of the two workflow runs would "lose": They would both try to delete the same resources, and the one who's on second place would then get a nasty error.

_Iff_ the need arises, we could add another `elif` that looks whether any competing `delete-self-hosted-runner` was queued or is running by executing the following query:

```
gh api repos/$GITHUB_REPOSITORY/actions/runs --jq '.workflow_runs[] | select(
    .path == ".github/workflows/delete-self-hosted-runner.yml"
    and .name == "Delete $vm_name"
    and (.status == "in_progress" or .status == "queued")
) | .id'
```

If we do this, we need to also add `GH_TOKEN: ${{ github.token }}` since unauthenticated REST API calls are notorious on GitHub Actions for being throttled by virtue of running on the same IP address as so many other unauthenticated REST API calls.

dscho:
Hrm. I looked something up in my Azure Portal and noticed some Windows/ARM64 being present. And it was the same that had been mentioned in [the cleanup run earlier today](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10871761622)!

So I [re-ran that cleanup workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10873127742), and it said the same thing: that the same VM was too old and needed to be deleted. But again, it wasn't!

Only when I triggered [a manual `delete-self-hosted-runner` workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10873144087) did the VM get deleted.

Any ideas @dennisameling?

dscho:
Oh, and I noticed something else entirely: [neither `node` nor `gh` are found](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10873127742/job/30168884992#step:5:51)!

I guess we need to fiddle with this workflow a bit more, after all. How about switching away from `azure/CLI` to a regular `run:` step? Azure CLI [is installed in the ubuntu-latest runners](https://github.com/actions/runner-images/blob/ce2a3e8e67858dc8c3dd77aca6b29e2956d934ed/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L105), after all.

dennisameling:
> How about switching away from `azure/CLI` to a regular `run:` step? Azure CLI [is installed in the ubuntu-latest runners](https://github.com/actions/runner-images/blob/ce2a3e8e67858dc8c3dd77aca6b29e2956d934ed/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L105), after all.

Ah, good catch! Will file a PR to fix soon.

dennisameling:
[Here's the follow-up PR](https://github.com/git-for-windows/git-for-windows-automation/pull/96).

PR #96, opened 2024-09-18
Self-hosted runner cleanup: use Ubuntu instead of azure/cli action

dennisameling:
The `azure/cli` GitHub Action runs in a Docker container with Alpine or Azure Linux, meaning that many tools like `node` and `gh` (which we rely on) aren't available.

Since the Azure CLI is available in the GitHub-hosted `ubuntu-latest` image anyway, let's leverage that instead. In the end, it has all the other tools we need as well.

Ref: https://github.com/actions/runner-images/blob/ce2a3e8e67858dc8c3dd77aca6b29e2956d934ed/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L105

### DIFF
--- a/.github/workflows/cleanup-self-hosted-runners.yml
+++ b/.github/workflows/cleanup-self-hosted-runners.yml
@@ -17,6 +17,7 @@ on:
 #                        group specifically for self-hosted Actions Runners.
 jobs:
   delete-runner:
+    if: github.repository_owner == 'git-for-windows'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/cleanup-self-hosted-runners.yml
+++ b/.github/workflows/cleanup-self-hosted-runners.yml
@@ -13,58 +13,78 @@ on:
 #   az ad sp create-for-rbac --name "{YOUR_DESCRIPTIVE_NAME_HERE}" --role contributor \
 #     --scopes /subscriptions/{SUBSCRIPTION_ID_HERE}/resourceGroups/{RESOURCE_GROUP_HERE} \
 #     --sdk-auth
-# AZURE_RESOURCE_GROUP - Resource group to create the runner(s) in
+# AZURE_RESOURCE_GROUP - Resource group to find the runner(s) in. It's recommended to set up a resource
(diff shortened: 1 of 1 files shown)


dennisameling:
@dscho I need some help here regarding the Azure Credentials. Currently, you have an `AZURE_CREDENTIALS` secret in this repo/org, which is most likely a JSON payload, [as mentioned here](https://github.com/marketplace/actions/azure-login#login-with-a-service-principal-secret). It'll look something like this:

```json
{
    "clientSecret":  "******",
    "subscriptionId":  "******",
    "tenantId":  "******",
    "clientId":  "******"
}
```

If we want to invoke `az login` directly, which I did in this PR, we'll need to pass all of these individually. I'd opt for having them in dedicated secrets, allowing us to use them like this:

```
az login --service-principal -u ${{ secrets.AZURE_CLIENT_ID }} -p ${{ secrets.AZURE_CLIENT_SECRET }} --tenant ${{ secrets.AZURE_TENANT_ID }}
az account set --subscription ${{ secrets.AZURE_SUBSCRIPTION_ID }}
```

Would it be possible for you to set up those secrets? If not (or if it's too much work to do), I can most likely just extract the individual values from the JSON using `jq`, but I'm worried that GitHub might not mask those individual extracted values in the logs unless we do an `::add-mask::{value}` for each individual value first. Having them in separate secrets just feels a bit cleaner and easier to work with to me, but it can be worked around for sure.

WDYT?

Also, here's a [sample run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10929556857/job/30340524967), which obviously failed because of the missing secrets.

dennisameling:
> How about using a `github-script` step to extract those values instead of using `jq`? That should take care of the masking very easily, as well as of setting the values as step outputs.

@dscho great suggestion! I've applied the changes. [Here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10981992183/job/30489711403)'s a successful workflow run. Starting [another workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/10982010035) confirmed that no VMs are left after the cleanup 👍🏼

dscho (approved):
Awesome!

PR #98, opened 2024-09-25
Azure self-hosted runners: show logs on failure

dennisameling:
We're [seeing](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/11025772033/job/30621189230) intermittent `OSProvisioningTimedOut` errors on VM creation, which typically happens within 20-40 mins of the deployment hanging. Most creation jobs seem to work fine, though.

Let's try and find out what's going wrong, so we can see if we need to fix things on our end.

### DIFF
--- a/.github/workflows/create-azure-self-hosted-runners.yml
+++ b/.github/workflows/create-azure-self-hosted-runners.yml
@@ -158,6 +158,22 @@ jobs:
         template: ./azure-self-hosted-runners/azure-arm-template.json
         parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json ${{ env.AZURE_ARM_PARAMETERS }}
         scope: resourcegroup
+
+    - name: Show some more information on failure
+      if: failure()
+      run: |
+        echo "::group::VM status"
+        az vm get-instance-view --resource-group ${{ secrets.AZURE_RESOURCE_GROUP }} --name ${{ steps.generate-vm-name.outputs.vm_name }} --query "instanceView.statuses"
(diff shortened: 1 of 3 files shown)


.github/workflows/create-azure-self-hosted-runners.yml
```diff
@@ -158,8 +158,28 @@ jobs:
         template: ./azure-self-hosted-runners/azure-arm-template.json
         parameters: ./azure-self-hosted-runners/azure-arm-template-example-parameters.json ${{ env.AZURE_ARM_PARAMETERS }}
         scope: resourcegroup
+
+    - name: Show some more information on failure
+      if: failure()
+      uses: azure/cli@v2
```
dscho, line 164: Elsewhere, we moved away from Azure/CLI, maybe we can do that here, too?

dscho (approved):
Great stuff! Let's do this.

dennisameling:
@dscho I've updated the PR to move away from the `@azure/cli` action. I've also created a Composite Action for the login part, which we can reuse. However, I just learned that this may not add much benefit compared to the `@azure/login` action. I was under the assumption that said action only worked in combination with Azure-provided actions, but turns out it simply runs `az login` and `az account set` on the OS [behind the scenes](https://github.com/Azure/login/blob/a514566bcd042073414e083c41a3239d2f0f881f/src/Cli/AzureCliLogin.ts#L101).

So I'd be OK to keep using `@azure/login` if we want (since we can also run our own `az` commands after that step), otherwise we could use our own Composite Action instead. Any preference from your side?

Also, here's a [successful workflow run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/11157468502/job/31011802840) with the new setup.

dscho (approved):
I like the Composite Action!

PR #100, opened 2024-11-19
release-git: add missing aarch64 input to github-release action

dennisameling:
In earlier commits, we added suppport for ARM64 to the release pipelines, but there's a couple more places to update. Let's do that.

### DIFF
--- a/.github/workflows/release-git.yml
+++ b/.github/workflows/release-git.yml
@@ -125,6 +125,7 @@ jobs:
           artifacts-token: ${{ secrets.GITHUB_TOKEN }}
           git_artifacts_i686_workflow_run_id: ${{ env.I686_WORKFLOW_RUN_ID }}
           git_artifacts_x86_64_workflow_run_id: ${{ env.X86_64_WORKFLOW_RUN_ID }}
+          git_artifacts_aarch64_workflow_run_id: ${{ env.AARCH64_WORKFLOW_RUN_ID }}
   announcement-mail:
     needs: ['setup', 'github-release']
     runs-on: ubuntu-latest
--- a/.github/workflows/release-git.yml
+++ b/.github/workflows/release-git.yml
@@ -189,6 +189,7 @@ jobs:
           artifacts-token: ${{ secrets.GITHUB_TOKEN }}
           git_artifacts_i686_workflow_run_id: ${{ env.I686_WORKFLOW_RUN_ID }}
           git_artifacts_x86_64_workflow_run_id: ${{ env.X86_64_WORKFLOW_RUN_ID }}
+          git_artifacts_aarch64_workflow_run_id: ${{ env.AARCH64_WORKFLOW_RUN_ID }}
           gpg-key: ${{ secrets.GPGKEY }}
           priv-gpg-key: ${{ secrets.PRIVGPGKEY }}
           azure-blobs-token: ${{ secrets.AZURE_BLOBS_TOKEN }}
--- a/.github/actions/pacman-packages/action.yml
+++ b/.github/actions/pacman-packages/action.yml
@@ -33,6 +33,9 @@ inputs:
   git_artifacts_x86_64_workflow_run_id:
     description: 'ID of the git-artifacts (x86_64) workflow run'
     required: true
+  git_artifacts_aarch64_workflow_run_id:
+    description: 'ID of the git-artifacts (aarch64) workflow run'
+    required: true
   gpg-key:
     description: 'The GPG key to use to sign the updated Pacman database'
     required: true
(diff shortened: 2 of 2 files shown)


.github/workflows/release-git.yml
```diff
@@ -125,6 +125,7 @@ jobs:
           artifacts-token: ${{ secrets.GITHUB_TOKEN }}
           git_artifacts_i686_workflow_run_id: ${{ env.I686_WORKFLOW_RUN_ID }}
           git_artifacts_x86_64_workflow_run_id: ${{ env.X86_64_WORKFLOW_RUN_ID }}
+          git_artifacts_aarch64_workflow_run_id: ${{ env.AARCH64_WORKFLOW_RUN_ID }}
```
dennisameling, line 128: @dscho I was just going through the entire chain again, and noticed that I missed this part to pass `git_artifacts_aarch64_workflow_run_id` to the workflow.

While adding this input, I got `Invalid action input 'git_artifacts_aarch64_workflow_run_id'` in VS Code. I'm pretty sure that's because line 114 uses `git-for-windows/git-for-windows-automation/.github/actions/github-release@release` (the `release` branch), which isn't up-to-date with the latest changes from the `main` branch yet. Is that intentional?

dennisameling:
I also have a question about the `pacman-packages` job in this pipeline. It's currently [designed](https://github.com/git-for-windows/git-for-windows-automation/blob/bea16329800cf42756827a6fab47acd73562031a/.github/actions/pacman-packages/action.yml#L116) to update the Pacman repos for `x86_64` and `i686` in one go, and I'm not sure whether I can simply add `aarch64` to it. While `please.sh quick_add` already supports `aarch64`, I saw in the past that everything `pacman`-related for ARM64 runs more reliably on a native ARM64 runner, e.g. post-install scripts.

Do you think that in this case, where we would be publishing to the `aarch64` pacman repo, we could get away with doing so on the existing x64 runner logic, or should I create a matrix to run the `aarch64` logic on a native ARM64 runner instead?

.github/workflows/release-git.yml
dscho: > While adding this input, I got `Invalid action input 'git_artifacts_aarch64_workflow_run_id'` in VS Code. I'm pretty sure that's because line 114 uses `git-for-windows/git-for-windows-automation/.github/actions/github-release@release` (the `release` branch), which isn't up-to-date with the latest changes from the `main` branch yet. Is that intentional?

Yes, this is intentional, and the `release-git` workflow tries to [always ensure that `main` and `release` start out identically](https://github.com/git-for-windows/git-for-windows-automation/blob/bea16329800cf42756827a6fab47acd73562031a/.github/workflows/release-git.yml#L37-L80).

0f5ecb3accc20fbd10e9d6c688da18ad9b0a8e3d explains the rationale a bit, and 060621ab8672ce51d78f67ed3cbc34ca45798da0 paints a fuller picture.

The tl;dr is: release automation is hard, and the idea is delusional that one could get it all right in a single YAML file that won't need to adapt to external changes, ever, or at least can be adapted correctly _before_ deploying it. Instead, the need to adapt to external changes are typically detected in the middle of a release, with half the things deployed, and the rest still needing to be deployed. Using the composite Actions from a _different_ branch allows _that_ branch to be updated and the workflow job can then be restarted _after editing just that job_.

Release automation is a lot easier in Azure DevOps' original design, without YAML. You can edit failed tasks in the web browser and then re-run from there. Highly convenient. But then there was this huge push for YAML a couple of years ago, and GitHub Actions was designed with that in mind and therefore only offers YAML _that must be part of the commit history_, that's why we're stuck with this ugly, ugly, _uggallee_ work-around. GitHub Actions simply lacks the support we need for complex deployments in the real world.

dscho:
> I also have a question about the `pacman-packages` job in this pipeline. It's currently [designed](https://github.com/git-for-windows/git-for-windows-automation/blob/bea16329800cf42756827a6fab47acd73562031a/.github/actions/pacman-packages/action.yml#L116) to update the Pacman repos for `x86_64` and `i686` in one go, and I'm not sure whether I can simply add `aarch64` to it.

I think it should be safe.

> While `please.sh quick_add` already supports `aarch64`, I saw in the past that everything `pacman`-related for ARM64 runs more reliably on a native ARM64 runner, e.g. post-install scripts.

That's when installing them. [`pacman-helper.sh quick_add](https://github.com/git-for-windows/build-extra/blob/5d189481c541e1a9af47d20592d6fb7d4b1e19aa/pacman-helper.sh#L544-L740)` does not install them.

In fact, I had tried to move the `pacman-packges` job to an Ubuntu runner (for speed), but it [calls `repo_add`](https://github.com/git-for-windows/build-extra/blob/5d189481c541e1a9af47d20592d6fb7d4b1e19aa/pacman-helper.sh#L686) (actually [twice](https://github.com/git-for-windows/build-extra/blob/5d189481c541e1a9af47d20592d6fb7d4b1e19aa/pacman-helper.sh#L696)), and that _requires_ at least a part of Git for Windows' SDK.

I haven't had time to play around with it yet, but we could even consider using [the custom composite `init-g4w-sdk-for-pacman` GitHub Action](https://github.com/git-for-windows/git-for-windows-automation/blob/main/.github/actions/init-g4w-sdk-for-pacman/action.yml) in the `pacman-packages` job (after adding `repo_add` to [the sparse-checkout definition](https://github.com/git-for-windows/git-for-windows-automation/blob/bea16329800cf42756827a6fab47acd73562031a/.github/actions/init-g4w-sdk-for-pacman/action.yml#L47-L56)).

> Do you think that in this case, where we would be publishing to the `aarch64` pacman repo, we could get away with doing so on the existing x64 runner logic, or should I create a matrix to run the `aarch64` logic on a native ARM64 runner instead?

I think that we can totally just add `aarch64` packages in the same job, on an x86_64 runner.

dscho:
> > I also have a question about the `pacman-packages` job in this pipeline. It's currently [designed](https://github.com/git-for-windows/git-for-windows-automation/blob/bea16329800cf42756827a6fab47acd73562031a/.github/actions/pacman-packages/action.yml#L116) to update the Pacman repos for `x86_64` and `i686` in one go, and I'm not sure whether I can simply add `aarch64` to it.
> 
> I think it should be safe.

Let's include this also in this here PR?

mjcheetham:
> Let's include this also in this here PR?

@dennisameling @dscho I'm happy to help here and push the required extra changes!

dscho (approved):

PR #101, opened 2024-11-26
self-hosted runners: use native arm64 version of GfW

dennisameling:
Closes #28 

2.47.1 is the first version of Git for Windows for which native arm64 binaries are being published. Let's use those moving forward, for better performance.

Ref: https://github.com/git-for-windows/git/releases/tag/v2.47.1.windows.1

PR #102, opened 2024-11-26
post-deployment-script: fix regex for sha256

dennisameling:
Follow-up of https://github.com/git-for-windows/git-for-windows-automation/pull/101

There was a bug in the regex that finds the sha256 hash corresponding to the Git binary. Doing things like `[Regex]::Escape` directly in a string interpolation, apparently isn't supported by PowerShell. It just returned an empty string in the string interpolation.

This has surprisingly worked until now, because `Git-2.47.1-64-bit.exe` was [always the first binary in the list of assets](https://github.com/git-for-windows/git/releases/tag/v2.47.1.windows.1), so the corresponding hash was found. Now that we're getting the native `-arm64.exe` binary, the bug in the logic got exposed.

This commit fixes the bug by assigning the escaped asset name to a dedicated variable first.

### DIFF
--- a/azure-self-hosted-runners/post-deployment-script.ps1
+++ b/azure-self-hosted-runners/post-deployment-script.ps1
@@ -52,7 +52,8 @@ Write-Output "Starting post-deployment script."
 try {
     [System.Object]$GithubRestData = Invoke-RestMethod -Uri $GitHubUrl -Method Get -Headers $GithubHeaders -TimeoutSec 10 | Select-Object -Property assets, body
     [System.Object]$GitHubAsset = $GithubRestData.assets | Where-Object { $_.name -match $GithubExeName }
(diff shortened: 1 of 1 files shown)


dscho (approved):
Sounds good.

For easier testing, do you think we could find a way to embed the script in the deployment configuration rather than having to have a central download URL?

rimrul:
> Hopefully, the entire self-hosted setup will no longer be necessary by the end of this year, which is when GitHub-hosted arm64 runners [should become available for open source projects](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/#get-started-using-arm-hosted-runners-today).

That has been moved to the first quarter of 2025


https://github.com/orgs/github/projects/4247/views/1?sliceBy%5BcolumnId%5D=&layout=roadmap&filterQuery=is%3Aopen+-status%3AFuture+arm&pane=issue&itemId=65252539&issue=github%7Croadmap%7C970

PR #104, opened 2025-01-04
llvm: add changed patches

rimrul:
As we noticed in https://github.com/git-for-windows/MINGW-packages/pull/132 the update script for llvm currently doesn't properly handle added or removed patches. This can break the llvm build. Update the script to handle these patches.


dscho (approved):
👍

PR #107, opened 2025-01-09
release-git: rank ARM64 higher than 32-bit

dscho:
The SHA-256 checksums of the Git for Windows release artifacts are listed at the bottom of Git for Windows' announcement emails.

Certain of those artifacts are deemed more important than others and are hence listed first. For example, x86_64 is nowadays considered much more important than i686 (which is deprecated, after all). To present the artifacts in such an order, the `artifactName2Rank()` function hard-codes CPU architecture rankings.

When ARM64 artifacts were included, this ranking was not adjusted, and as a consequence ARM64 ranked way on the bottom. However, ARM64 is the future (or will be, once GitHub comes through with the promise of hosted runners) because of battery life, performance potentials, etc, and it should therefore outrank i686.

Let's make it so!

### DIFF
--- a/github-release.js
+++ b/github-release.js
@@ -92,8 +92,8 @@ const ranked = artifacts
 const artifactName2Rank = (name) => {
   let rank = ranked.indexOf(name
     .replace(/-\d+(\.\d+)*(-rc\d+)?/, '')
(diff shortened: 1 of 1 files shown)


mjcheetham (approved):
Make it so.

PR #109, opened 2025-01-27
Start migrating Git for Windows' snapshot builds to GitHub Actions

dscho:
In 2017, I started working on [the `GitArtifacts` Azure Pipeline](https://dev.azure.com/git-for-windows/git/_build?definitionId=34&_a=summary) to not only produce Git for Windows' artifacts (such as installer, portable Git, MinGit, etc) for _full_ releases, but basically on every push. The resulting artifacts were uploaded to Azure Blobs and for convenience, an index of all of those snapshots was maintained at https://wingit.blob.core.windows.net/files/index.html.

This all depends on Azure Blob storage that is hosted in my personal Azure account.

I recently stumbled over [yet another excellent letter to open-source maintainers](https://xuanwo.io/2024/10-a-letter-to-open-source-maintainers/) that had this great [piece](https://xuanwo.io/2024/10-a-letter-to-open-source-maintainers/#always-ready-to-leave) of advice:

> Essential workflows, such as release processes and integration tests, should not rely on your personal account.

In an attempt to start remedying this, I have transferred all of the snapshot artifacts to releases at https://github.com/git-for-windows/git-snapshots/releases/, and moved the index to https://gitforwindows.org/git-snapshots/.

Then, I triggered a [`tag-git` run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12965286926), which upon completion (thanks to [the GitForWindowsHelper GitHub App](https://github.com/git-for-windows/gfw-helper-github-app/pull/29)) triggered a new set of `git-artifacts` runs ([i686](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12965293791), [x86_64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12965293791) and [aarch64](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12965294344)). There was a little bit of a hiccup in the previously un-tested snapshots-only part of the i686 run that uncovered a really old bug (which I am trying to fix in https://github.com/gitgitgadget/git/pull/1857) which I had to [work around](https://github.com/git-for-windows/git-for-windows-automation/commit/fe95907cf7fdb9babfc7ea719f5753bd01daf6ce) and then trigger [a new i686 run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/12969105876).

After that, I finally could start working on a new GitHub workflow called `upload-snapshot`, which is presented here. It takes the three run IDs of the i686, x86_64 and aarch64 workflow runs as input and takes it from there.

As a bonus of moving this process over from Azure Pipelines, where we enjoy even less support for Windows/ARM64 runners than in GitHub Actions, snapshots now also have ARM64 artifacts.

Note that I used some functionality that I had to develop during the recent embargoed Git release phase (because a lot of functionality is missing from GitHub Actions and from GitHub repositories that are needed to do embargoed releases, for example, you cannot keep workflow runs private until a given publication date, the only way to do private builds is to have them done in private repositories, Actions do not work in the private forks associated with security advisories either, and it is an unfortunate consequence that there is no way to move the involved private builds to the public repository; Also, most techniques used in public repositories won't Just Work™️ in private repositories because you have to pass around tokens all of a sudden. Just to name a few complications), and I used the opportunity to sneak them into this here repository (i.e. the public one, not used for embargoed releases, but I definitely want the code here for transparency). There are still quite some patches left, but I have to polish them first so that they don't break _non-embargoed_ releases.

Do also note that this PR is only half of what is required to fully port the Git for Windows snapshots process to GitHub Actions. Since there is no way to trigger workflow runs on pushes to a different repository than where the workflow runs live (yet another thing I miss from Azure Pipelines), the GitForWindowsHelper GitHub App will need to be taught to run the `tag-git` workflow on `push` *unless* there is already a `tag-git` check run attached to the `git-for-windows/git` commit. (Note that the difference between workflow run and check run is quite important in this context.) If there already is one, then the `upload-snapshot` workflow needs to be triggered (uploading already-released Git for Windows versions, which is the scenario where a `tag-git` check run already exists). *And* GitForWindowsHelper also needs to learn to react to completed `git-artifacts` *snapshot* check runs on `git-for-windows/git`, triggering the `upload-snapshot` workflow also in that case. It needs to be careful, though, to verify that this is actually wanted, maybe by having a peek at `git-for-windows/git-artifacts`' releases to see whether a corresponding tag already exists, and maybe by doing an `ahead` check to see whether the corresponding commit is actually reachable from `git-for-windows/git`'s `main` branch.

### DIFF
--- a/prepare-embargoed-branches.sh
+++ b/prepare-embargoed-branches.sh
@@ -1,0 +1,116 @@
+#!/bin/sh
+
+die () {
+  echo "$*" >&2
+  exit 1
+}
+
+dry_run=
+mingit=
+while case "$1" in
+--dry-run|-n) dry_run=1;;
+--mingit) mingit=1;;
+-*) die "Unknown option: $1";;
+*) break;;
+esac; do shift; done
+
+test $# = 1 ||
+die "Usage: $0 [--dry-run] [--mingit] <version> # e.g. 2.39.1"
+
+version=${1#v}
+if test -z "$mingit"
+then
+	case "$version" in
+	*.*.*.windows.*)
+		# major.minor.patch.windows.extra
+		previous_version_prefix=${version%.windows.*}
+		version="${version%.windows.*}.${version##*.windows.}"
+		;;
+	*.*.*\(*)
+		# major.minor.patch(extra)
+		previous_version_prefix=${version%(*}
+		version="${version%(*}.${version##*(}"
+		version=${version%)}
+		;;
+	*[^0-9.]*|*..*|.*|*.) die "Invalid version: '$version'";;
+	*.*.*.*)
+		# major.minor.patch.extra
+		v0="${version#*.*.*.}"
+		previous_version_prefix=${version%.$v0}
+		;;
+	*.*.*) previous_version_prefix=${version%.*}.$((${version##*.}-1));; # major.minor.patch
+	*) die "Invalid version: '$version'";;
+	esac
+	branch_name=git-$version
+else
+	previous_version_prefix="$(expr "$version" : '\([0-9]\+\.[0-9]\+\)\.\{0,1\}[0-9]*$')"
+	test -n "$previous_version_prefix" || die "Invalid version: '$version'"
+	branch_name=mingit-$previous_version_prefix.x-releases
+fi
+grep_version_regex="$(echo "$previous_version_prefix" | sed 's/\./\\\\&/g')"
+
+handle_repo () {
+	name="$1"
+	path="$2"
+	args="$3"
+
+	echo "### Handling $name ###" &&
+
+	if test -e "$path/.git"
+	then
+		git_dir="$path/.git"
+		main_refspec="refs/remotes/origin/main:refs/heads/main"
+	else
+		# To allow for running this script on Linux/macOS, fall back to cloning to pwd
+		git_dir=${path##*/}.git &&
+		if test ! -d "$git_dir"
+		then
+			# We only need a partial clone
+			git clone --bare --filter=blob:none \
+				https://github.com/git-for-windows/$name "$git_dir"
+		fi
+		main_refspec="refs/heads/main:refs/heads/main"
+	fi &&
+
+	# ensure that the `embargoed-git-for-windows-builds` remote is set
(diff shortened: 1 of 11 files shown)


prepare-embargoed-branches.sh
```diff
@@ -0,0 +1,99 @@
...
+			echo "git push embargoed-git-for-windows-builds $push_ref_spec" &&
+			git --git-dir "$git_dir" push embargoed-git-for-windows-builds $push_ref_spec
+		fi
+	fi
+}
+
+handle_repo git-sdk-32 /c/git-sdk-32 \
+	"\"--grep=mingw-w64-i686-git \".*\" -> $grep_version_regex\" -- cmd/git.exe" &&
+handle_repo git-sdk-64 /c/git-sdk-64 \
+	"\"--grep=mingw-w64-x86_64-git \".*\" -> $grep_version_regex\" -- cmd/git.exe" &&
```
mjcheetham, line 110: Nothing for the ARM64 SDK?

prepare-embargoed-branches.sh
```diff
@@ -93,6 +93,8 @@ handle_repo git-sdk-32 /c/git-sdk-32 \
 	"\"--grep=mingw-w64-i686-git \".*\" -> $grep_version_regex\" -- cmd/git.exe" &&
 handle_repo git-sdk-64 /c/git-sdk-64 \
 	"\"--grep=mingw-w64-x86_64-git \".*\" -> $grep_version_regex\" -- cmd/git.exe" &&
+handle_repo git-sdk-arm64 /c/git-sdk-arm64 \
```
mjcheetham, line 111: And here it is.. never mind! :)

mjcheetham (approved):
Thanks for the thorough description and splitting the commits into small pieces :)

prepare-embargoed-branches.sh
dscho: I kept it a _little_ closer to the original, which was actually developed during the _previous_ embargo, the one leading up to v2.45.1... and back then, we indeed had no official support for Windows/ARM64 and hence that part was not needed in the embargoed branches.

PR #113, opened 2025-02-17
Start transitioning to Git for Windows' new Pacman repository

dscho:
To disentangle Git for Windows' Pacman repository from my personal Azure account a bit, I started transitioning it to https://github.com/git-for-windows/pacman-repo.

In https://github.com/git-for-windows/build-extra/pull/595, I prepare `pacman-helper.sh` (which is Git for Windows' work horse for Pacman package deployments) to dual-deploy both to the old and to the new Pacman repository. Naturally, this needs the appropriate GitHub token to work. So here is the PR to adjust _the_ two remaining call sites in Git for Windows' automation to provide said token.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -306,22 +306,13 @@ jobs:
           name: artifacts
           path: artifacts
 
-      - name: Prepare home directory for deploying to the Pacman repository
-        if: env.BUILD_ONLY != 'true' && env.AZURE_BLOBS_TOKEN != ''
-        env:
-          AZURE_BLOBS_TOKEN: ${{secrets.AZURE_BLOBS_TOKEN}}
-        shell: bash
-        run: |
-          echo "::add-mask::$(echo "$AZURE_BLOBS_TOKEN" | base64 -w 0)" &&
-          echo "$AZURE_BLOBS_TOKEN" >"$HOME"/.azure-blobs-token
-
       - name: ${{ env.BUILD_ONLY == 'true' && 'Test-deploy' || 'Deploy' }} Pacman packages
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -306,22 +306,22 @@ jobs:
...
 
       - name: ${{ env.BUILD_ONLY == 'true' && 'Test-deploy' || 'Deploy' }} Pacman packages
-        if: env.BUILD_ONLY == 'true' || env.AZURE_BLOBS_TOKEN != ''
+        if: env.BUILD_ONLY == 'true' || env.azure_blobs_token != ''
         shell: bash
         env:
           GPGKEY: ${{secrets.GPGKEY}}
           PACMANDRYRUN: ${{env.BUILD_ONLY}}
-          AZURE_BLOBS_TOKEN: ${{secrets.AZURE_BLOBS_TOKEN}}
+          azure_blobs_token: ${{secrets.AZURE_BLOBS_TOKEN}}
```
mjcheetham, line 323: Only just realised.. why the change to lowercase for the environment variable here?

dscho: That's because that is the spelling of [the environment variable, which, if unset, is set to the contents of `~/.azure-blobs-token`](https://github.com/git-for-windows/build-extra/blob/9518f2c4c365814e793fdf597e2fed988a028c0e/pacman-helper.sh#L710-L715). And if neither the (lower-case-keyed) environment variable nor that file exist, the deployment will fail.

And yes, Bash makes a difference between upper-case and lower-case environment variable names. That's why I can get away with setting `path` to a filename (and not royally mess up `PATH` while doing so).

mjcheetham (approved):

PR #117, opened 2025-02-25
git-artifacts: stop building i686 artifacts except MinGit

dscho:
As announced multiple times and also documented in detail at https://gitforwindows.org/32-bit, Git for Windows will no longer publish 32-bit installers, portable Gits and such.

The only i686 artifacts that will be published are the MinGit ones, until April 2029.

This addresses https://github.com/git-for-windows/git/issues/5394.

### DIFF
--- a/create-artifacts-matrix.js
+++ b/create-artifacts-matrix.js
@@ -7,26 +7,27 @@ module.exports = (artifactsString, architecture) => {
 
     const validArtifacts = [
         {
-            name: 'installer',
-            filePrefix: 'Git',
-            fileExtension: 'exe'
-        },
-        {
-            name: 'portable',
-            filePrefix: 'PortableGit',
-            fileExtension: 'exe'
-        },
-        {
-            name: 'archive',
-            filePrefix: 'Git',
-            fileExtension: 'tar.bz2'
-        },
-        {
             name: 'mingit',
             filePrefix: 'MinGit',
             fileExtension: 'zip'
         }
     ]
+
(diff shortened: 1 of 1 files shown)


rimrul (approved):

mjcheetham (approved):

dennisameling:
Do you see any value in starting to already remove i686-related logic like [here](https://github.com/git-for-windows/build-extra/blob/68def8c7151bc7af622679dcb7b22512e0d83fa9/portable/release.sh#L41)? I might be able to help a bit with that in the next couple weeks, but I'm sure there's more pressing things to address first.

dscho:
> Do you see any value in starting to already remove i686-related logic like [here](https://github.com/git-for-windows/build-extra/blob/68def8c7151bc7af622679dcb7b22512e0d83fa9/portable/release.sh#L41)?

There is nothing urgent about this, we have plenty of time, and if anything, I would first want to concentrate on https://github.com/git-for-windows/git/issues/5401, which will require a change in how the i686 MINGW packages are built in [`build-and-deploy`](https://github.com/git-for-windows/git-for-windows-automation/actions/workflows/build-and-deploy.yml): We will want to build the i686 packages in a separate workflow than the x86_64, so that the former can run exclusively in a `git-sdk-32` and the latter exclusively in a `git-sdk-64` environment.

> I might be able to help a bit with that in the next couple weeks, but I'm sure there's more pressing things to address first.

That is very kind of you! But as I said, nothing about this is urgent.

PR #129, opened 2025-05-05
Prepare for i686-only MINGW builds

dscho:
In the context of https://github.com/git-for-windows/git/issues/5401 ("Drop all i686 stuff from git-sdk-64"), this PR adjusts a few files so that the `build-and-deploy` workflow will run successfully in `git-sdk-32` (it currently runs in `git-sdk-64`, unlike when the `git-artifacts` builds `mingw-w64-i686-git`, where `git-sdk-32` is used already).

The `build-and-deploy` workflow will continue as before, as it will only make a difference when the `architecture` is set to `i686` or `x86_64` to build a MINGW package (and currently, the `architecture` is left empty for those).

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -8,7 +8,7 @@ on:
         description: 'The package to build'
         required: true
       build_only:
-        description: 'Skip deployment (if non-empty)'
+        description: 'Skip deployment (if "true")'
         required: false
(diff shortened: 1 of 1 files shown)


dscho:
[Here](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/16906401657/job/47897092927) is a test run for i686 `mingw-w64-curl`.

PR #130, opened 2025-05-22
drop support for selfhosted runners

rimrul:
We're using GitHub hosted runners nowadays and don't need to manage our own ARM64 runners anymore.


dennisameling (approved):
Thanks! We might want to wait a little bit longer to merge this until the hosted runners hit GA, but in case any problems arise with those, we can also revert this here commit that deleted the self-hosted runner code. 

WDYT?

rimrul:
> We might want to wait a little bit longer to merge this until the hosted runners hit GA,

Looking at the `ubuntu-22.04-arm` image, it's been [in public beta for almost a year](https://github.com/actions/partner-runner-images/commit/ed78002f4892088dd5c0c4dcc8b27b3089c3143d), so that could take a little while. And GA on partner runner images  doesn't offer additional guarantees compared to beta.

But I guess it makes sense to at least wait until we see how the early june update of the image turns out.

dscho:
I like the "let's wait and see" approach. We had to wait quite a while for the public preview, with many a delay, and the recent engineer-heavy cuts at GitHub may also well have their impact on initiatives unrelated to AI, like Windows/ARM64 runners GA.

PR #131, opened 2025-07-11
prepare-embargoed-branches: special-case some older release trains

dscho:
I had to release some security bug fix releases recently that did not _quite_ follow a previous version. For example, there was a gap between [Git for Windows v2.47.1(2)](https://github.com/git-for-windows/git/releases/tag/v2.47.1.windows.2) and [MinGit v2.47.3](https://github.com/git-for-windows/git/releases/tag/v2.47.3.windows.1) (there was no Git for Windows nor MinGit v2.47.2).

To accommodate for this, I came up with these patches (and used them in the private repository where the embargoed releases were built).


prepare-embargoed-branches.sh
```diff
@@ -47,6 +51,18 @@ else
...
 	branch_name=mingit-$previous_version_prefix.x-releases
 fi
+case "$previous_version" in
+2.46.3)
+	# There was no Git for Windows v2.46.3, so we use v2.46.2
+	previous_version=2.46.2
+	previous_version_prefix=2.46.2
+	;;
+2.47.2)
+	# There was no Git for Windows v2.47.3, so we use v2.47.1(2)
```
rimrul, line 61: /2.47.3/2.47.2/

dscho: Actually, no... This `switch` case is about the _previous_ version, that is, the script is called with `v2.47.3` and `previous_version` will then be set to `2.47.2`...

rimrul: Yes, the switch seems to be correct, but the comment claims that 2.47.3 didn't exist, whereas the code special cases 2.47.2 not existing.

PR #165, opened 2026-04-24
git-artifacts: switch code signing to Azure Artifact Signing with OIDC

dscho:
The current code signing certificate (used via `osslsigncode` and build-extra's `signtool.sh`) is expiring very soon. Unfortunately, renewing the certificate would be prohibitively expensive. There was a price hike (from $247.50 for three years to $345 per year), mainly due to the fact that now this code-signing method would require a hardware dongle, which incidentally is also incompatible with the way we currently build: we cannot just plug in a USB dongle into the cloud.

This PR switches to [Azure Artifact Signing](https://learn.microsoft.com/en-us/azure/trusted-signing/overview) using the [`dotnet/sign`](https://github.com/dotnet/sign) CLI, authenticating via OIDC rather than static secrets, nicely side-stepping both blockers.

### Why Azure Artifact Signing?

Cloud-managed certificates eliminate the need to handle private key material in CI. The signing key lives in an Hardware Security Module ("HSM"), and access is controlled through Azure Role-Based Access Control ("RBAC").

Concretely, this means that the authorization is keyed on the GitHub workflow running on the `main` branch. Even if a secret should leak, the blast radius is therefore very small.

### Why the pre-built sign tool?

The `dotnet/sign` project currently lacks native ARM64 support ([dotnet/sign#852](https://github.com/dotnet/sign/issues/852)). A [pre-built x64 `sign.exe`](https://github.com/dscho/prebuilt-dotnet-sign-tool/releases) (with bundled .NET runtime) runs fine under emulation on `windows-11-arm` runners, covering all three architectures from one tool.

### What changes

The new `azure-signtool.sh` is a drop-in replacement for build-extra's `signtool.sh`. It auto-downloads the sign tool on first use and preserves the same retry/backoff logic. The `git signtool` alias is pointed at this wrapper, so `please.sh` and the explicit Portable Git signing call work without further changes.

Authentication uses [OIDC workload identity federation](https://docs.github.com/en/actions/concepts/security/openid-connect) (`azure/login@v3` + `--azure-credential-type azure-cli`) instead of a static client secret. This means no long-lived credentials are stored in GitHub secrets at all; Azure only trusts short-lived tokens from this specific repository.

### Required secrets

| Secret | Purpose |
|---|---|
| `AZURE_CLIENT_ID` | App registration's Application (client) ID |
| `AZURE_TENANT_ID` | Azure AD tenant ID |
| `AZURE_SIGNING_OPTS` | Sign tool arguments, e.g. `--artifact-signing-endpoint <url> --artifact-signing-account <acct> --artifact-signing-certificate-profile <profile>` |

The Azure-side setup (app registration, federated credential, RBAC role assignment) is documented in the second commit's message, and I have already confirmed that it works, via [this `git-artifacts` run](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/24876848095) (which did pick up a previously-built `mingw-64-git` package, but did code-sign the installer with my new certificate).

### DIFF
--- a/azure-signtool.sh
+++ b/azure-signtool.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+
+# Azure Artifact Signing wrapper for use as `git signtool`.
+#
+# This replaces the osslsigncode-based signtool.sh from build-extra,
+# using the pre-built sign tool from
+# https://github.com/dscho/prebuilt-dotnet-sign-tool instead.
+#
+# It expects the following environment variables:
+#   AZURE_CLIENT_ID                      - Azure AD app client ID
+#   AZURE_CLIENT_SECRET                  - Azure AD app client secret
+#   AZURE_TENANT_ID                      - Azure AD tenant ID
+#   AZURE_SIGNING_ENDPOINT               - Azure Artifact Signing endpoint URL
+#   AZURE_SIGNING_ACCOUNT                - Azure Artifact Signing account name
+#   AZURE_SIGNING_CERTIFICATE_PROFILE    - certificate profile name
+
+die () {
+	echo "$*" >&2
+	exit 1
+}
+
+# Locate the sign tool, downloading it if necessary
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SIGN_TOOL_DIR="$SCRIPT_DIR/.sign-tool"
+sign_exe="$SIGN_TOOL_DIR/sign"
+
+if ! test -x "$sign_exe" && ! test -x "$sign_exe.exe"
+then
+	echo "Downloading pre-built sign tool..." >&2
+	sysdir="${SYSTEMROOT:-${WINDIR:-/proc/fake-non-existent}}/System32" &&
+	mkdir -p "$SIGN_TOOL_DIR" &&
+	url="https://github.com/dscho/prebuilt-dotnet-sign-tool/releases/latest/download/sign-tool.zip" &&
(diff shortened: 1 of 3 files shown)


azure-signtool.sh
```diff
@@ -0,0 +1,66 @@
...
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+SIGN_TOOL_DIR="$SCRIPT_DIR/.sign-tool"
+sign_exe="$SIGN_TOOL_DIR/sign"
+
+if ! test -x "$sign_exe" && ! test -x "$sign_exe.exe"
+then
+	echo "Downloading pre-built sign tool..." >&2
+	sysdir="${SYSTEMROOT:-${WINDIR:-/proc/fake-non-existent}}/System32" &&
+	mkdir -p "$SIGN_TOOL_DIR" &&
+	url="https://github.com/dscho/prebuilt-dotnet-sign-tool/releases/latest/download/sign-tool.zip" &&
```
mjcheetham, line 32: Do we want to eventually move this to the GfW org?

mjcheetham (approved):

azure-signtool.sh
dscho: I hope that that won't be necessary, as I would like to believe that proper `dotnet sign` support for Windows/ARM64 is not _too_ far off in the future.

But yes, if this should become a semi-permanent prerequisite, we should eventually move it.

PR #181, opened 2026-06-22
upload-snapshot: include UCRT64 artifacts when present

dscho:
Now that the git-artifacts workflow can build UCRT64 variants, this teaches `upload-snapshot.yml` to pick them up for the snapshots. UCRT64 is deliberately kept out of the full Git for Windows releases for the time being, so the pickup is lenient: a new optional `git_artifacts_ucrt64_workflow_run_id` input (with a matching `UCRT64_WORKFLOW_RUN_ID` environment variable) is honored only when it is provided, and even then only the UCRT64 artifacts that the run actually published are downloaded. A snapshot upload that runs against builds without UCRT64 (such as a full release, where this workflow picks up the bits that were produced) therefore contributes nothing extra.

The full-release path is left untouched: the `architectures` list in `github-release.js` that drives the release assembly keeps its three entries (`x86_64`, `i686`, `aarch64`), so full releases remain UCRT64-free.

Note that, to actually feed UCRT64 into the snapshots, whatever dispatches `upload-snapshot.yml` (the GitForWindowsHelper app) will need to pass the new `git_artifacts_ucrt64_workflow_run_id` input.

### DIFF
--- a/.github/workflows/upload-snapshot.yml
+++ b/.github/workflows/upload-snapshot.yml
@@ -13,6 +13,9 @@ on:
       git_artifacts_aarch64_workflow_run_id:
         description: 'ID of the git-artifacts (aarch64) workflow run'
         required: true
+      git_artifacts_ucrt64_workflow_run_id:
+        description: 'ID of the git-artifacts (ucrt64) workflow run (optional; UCRT64 is only included in snapshots)'
+        required: false
 
 env:
   OWNER: "${{ github.repository_owner }}"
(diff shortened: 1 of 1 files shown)


dscho:
Since the automation [looks for an existing `tag-git` run upon push to git-for-windows/git's `main` branch](https://github.com/git-for-windows/gfw-helper-github-app/blob/7af256b6f2c9ef0a04050706d877056bd002d6a3/GitForWindowsHelper/cascading-runs.js#L286), it will pick up even the UCRT64 bits for full releases, once https://github.com/git-for-windows/gfw-helper-github-app/pull/223 is merged.

PR #191, opened 2026-07-20
Accommodate longer GitHub App access tokens

dscho:
As per https://github.blog/?p=95566, the access tokens' length has been increased. This causes issues when constructing HTTP Authorization headers using `base64` without taking great care to avoid line-wrapping.

While working on this, I also noticed that some obsolete logic (that also calls `base64`) needed to be removed: There was still logic to deploy Pacman packages to Azure Blob storage, and old code-signing logic would be broken now because the pure file-based code-signing certificate has expired months ago (I only replaced that logic with Azure Artifact Signing-based logic for the `mingw-w64-git` package as part of the release process).

This is a companion of https://github.com/git-for-windows/build-extra/pull/724.


mjcheetham (approved):

.github/workflows/build-and-deploy.yml
```diff
@@ -194,19 +194,6 @@ jobs:
...
-          CODESIGN_P12: ${{secrets.CODESIGN_P12}}
-          CODESIGN_PASS: ${{secrets.CODESIGN_PASS}}
-        shell: bash
-        run: |
-          mkdir -p "$HOME"/.sig &&
-          echo "$CODESIGN_P12" | tr % '\n' | base64 -d >"$HOME"/.sig/codesign.p12 &&
-          echo "$CODESIGN_PASS" >"$HOME"/.sig/codesign.pass
-          git config --global alias.signtool '!sh "/usr/src/build-extra/signtool.sh"'
-          echo "SIGNTOOL=git signtool" >>$GITHUB_ENV
-
```
dscho, line 209: Whoopsie. I missed that [Bash was still code-signed](https://github.com/git-for-windows/git-for-windows-automation/actions/runs/27460744742/job/81173916591#step:15:3382), even though the signature is invalid (because the code-signing certificate had expired at that stage...).

